PR #2, opened 2024-01-30
Initial import

mimaison:
Copied source from the [PoC repository](https://github.com/mimaison/prometheus-metrics-reporter).

Changes made:
- updated license headers
- changed source packages to `io.strimzi.kafka.metrics`
- bumped some dependencies
- added `.gitignore`
- added `scm`, `issueManagement` and `license` sections to `pom.xml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ mvn package assembly:single
 ## Run
 
 ### Kafka Brokers
+
 Add the following to your broker configuration:
 ```properties
 metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ kafka.metrics.reporters=io.strimzi.kafka.metrics.YammerPrometheusMetricsReporter
 ```
 
 ### Kafka Clients
+
 Add the following to your client configuration:
 ```properties
 metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter
 
 ## Access Metrics
 
-Metrics are exposed on [http://localhost:8080/metrics](http://localhost:8080/metrics)
+Metrics are exposed on `http://localhost:8080/metrics`.
 
 ## Getting help
 
--- a/README.md
+++ b/README.md
@@ -6,6 +6,31 @@
 This repository contains the _Prometheus Metrics Reporter for Apache Kafka server and client components_ as proposed in [Strimzi Proposal #64](https://github.com/strimzi/proposals/blob/main/064-prometheus-metrics-reporter.md).
 The implementation is currently still in progress.
 
+## Build
+
+```shell
+mvn package assembly:single
+```
+
+## Run
+
+### Kafka Brokers
+Add the following to your broker configuration:
+```properties
+metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter
+kafka.metrics.reporters=io.strimzi.kafka.metrics.YammerPrometheusMetricsReporter
+```
+
+### Kafka Clients
+Add the following to your client configuration:
+```properties
+metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter
+```
+
+## Access Metrics
+
+Metrics are exposed on [http://localhost:8080/metrics](http://localhost:8080/metrics)
+
 ## Getting help
 
 If you encounter any issues while using Strimzi, you can get help using:
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,143 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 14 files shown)


README.md
```diff
@@ -6,6 +6,31 @@
...
+
+### Kafka Clients
+Add the following to your client configuration:
+```properties
+metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter
+```
+
+## Access Metrics
+
+Metrics are exposed on [http://localhost:8080/metrics](http://localhost:8080/metrics)
```
scholzj, line 32: Not sure if it really makes sense to make this a link as it will not work in most cases ... maybe formatting it as code would be better?

```suggestion
Metrics are exposed on `http://localhost:8080/metrics`.
```

scholzj (approved):

see-quick (approved):
LGTM, just one note to consider. I would add a simple JavaDoc for each class explaining its purpose.

README.md
```diff
@@ -6,6 +6,31 @@
...
+## Build
+
+```shell
+mvn package assembly:single
+```
+
+## Run
+
+### Kafka Brokers
+Add the following to your broker configuration:
```
ppatierno, line 18:
```suggestion

Add the following to your broker configuration:
```

README.md
```diff
@@ -6,6 +6,31 @@
...
+```properties
+metric.reporters=io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter
+kafka.metrics.reporters=io.strimzi.kafka.metrics.YammerPrometheusMetricsReporter
+```
+
+### Kafka Clients
+Add the following to your client configuration:
```
ppatierno, line 25:
```suggestion

Add the following to your client configuration:
```

ppatierno (approved):
LGTM. Just a couple of nits and agree with Maros about adding more Javadoc on classes.

mimaison:
Thanks for the feedback!

I added a short javadoc to all main classes. It's definitively something that will need to be improved.

pom.xml
```diff
@@ -0,0 +1,143 @@
...
+    </dependency>
+  </dependencies>
+
+  <build>
+    <pluginManagement><!-- lock down plugins versions to avoid using Maven defaults (may be moved to parent pom) -->
+      <plugins>
+        <!-- clean lifecycle, see https://maven.apache.org/ref/current/maven-core/lifecycles.html#clean_Lifecycle -->
+        <plugin>
+          <artifactId>maven-clean-plugin</artifactId>
+          <version>3.1.0</version>
```
Frawless, line 106: Does it make sense to put this version to `properties` above and just reference it here? Same for all other versions used withing `build`. It seems we do it in operator repo.

scholzj:
I think we should get it merged as the initial import ... and then work on the further changes in a separate PRs. We will anyway need changes to the Maven build to publish the builds, Javadocs for that as well etc.

Frawless (approved):

scholzj:
I think ghanging this line: https://github.com/strimzi/strimzi.github.io/blob/b5acaea8b7fab22b5cf54d5360692aa4b7378f73/_sass/strimzi.scss#L25 to `justify-content: center` will align it to center.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public List<MetricFamilySamples> collect() {
+        List<MetricFamilySamples> samples = new ArrayList<>();
+
+        for (Map.Entry<MetricName, KafkaMetric> entry : metrics.entrySet()) {
+            MetricName metricName = entry.getKey();
+            KafkaMetric kafkaMetric = entry.getValue();
+            LOG.trace("Collecting Kafka metric {}", metricName);
+
+            String name = metricName(metricName);
```
tombentley, line 49: Is there a reason we can't compute the name when inserting and removing metrics from the map, rather than here (which I assume is a hotter path)? If there is a reason it would be worth a comment.

src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
```diff
@@ -0,0 +1,148 @@
...
+                metricName.getGroup() + '_' +
+                metricName.getType() + '_' +
+                metricName.getName()).toLowerCase();
+        LOG.info("metricName group {}, type {}, name {} converted into {}", metricName.getGroup(), metricName.getType(), metricName.getName(), metricNameStr);
+        return metricNameStr;
+    }
+
+    static Map<String, String> labelsFromScope(String scope) {
+        if (scope != null) {
+            String[] parts = scope.split("\\.");
```
tombentley, line 99: We're paying the cost of compiling this regex with every call. It's probably worth pulling out and reusing a `Pattern`

tombentley (approved):
I left a couple of comments, but am quite happy for them to be fixed in later PRs. I know there's a bit of work to do to really flesh out the PoC.

kyguy (approved):
Nothing to add, great work Michael!

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
mimaison: The only reason it's here is to follow the same logic as in `YammerMetricsCollector`. But yeah for this code path this logic should be done when adding the metric instead of recomputing it each time the metrics are collected.

src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
mimaison: Yes, that's a potential performance improvement we should try

Issue #3, opened 2024-01-30
Change port configuration to listener


mimaison:
Currently the reporters have a `port` configuration. As per [the proposal](https://github.com/strimzi/proposals/blob/main/064-prometheus-metrics-reporter.md), the reporters should expose a `listener` configuration:
```
prometheus.metrics.reporter.listener: The listener to expose the metrics in the format http://<HOST>:<PORT>. If the <HOST> part if empty the listener binds to the default interface, if it is set to 0.0.0.0, the listener binds to all interfaces. If the <PORT> part is set to 0, a random port is picked. Default: http://:8080.
```

mimaison:
Done in https://github.com/strimzi/metrics-reporter/pull/16

PR #16, opened 2024-02-13
Change port configuration to listener

OwenCorrigan76:
This PR deals with the issue: https://github.com/strimzi/metrics-reporter/issues/3

### DIFF
--- a/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
+++ b/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
@@ -81,7 +81,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
                 int port = Integer.parseInt(matcher.group(2));
                 return new Listener(host, port);
             } else {
-                throw new ConfigException(LISTENER_CONFIG + ": " +  listener +  " Listener be of format http:// [host]:[port]");
+                throw new ConfigException(LISTENER_CONFIG, listener, "Listener be of format http:// [host]:[port]");
             }
         }
 
@@ -132,7 +132,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
         public void ensureValid(String name, Object value) {
             Matcher matcher = Listener.PATTERN.matcher(String.valueOf(value));
             if (!matcher.matches()) {
-                throw new ConfigException(name + ": "  + value + " Listener must be of format http:// [host]:[port]");
+                throw new ConfigException(name, value, "Listener must be of format http:// [host]:[port]");
             }
         }
     }
--- a/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
+++ b/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
@@ -76,12 +76,12 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
 
         public static Listener parseListener(String listener) {
             Matcher matcher = PATTERN.matcher(listener);
-            if (matcher.find()) {
+            if (matcher.matches()) {
                 String host = matcher.group(1);
                 int port = Integer.parseInt(matcher.group(2));
                 return new Listener(host, port);
             } else {
-                throw new ConfigException("Invalid listener format: " + listener);
+                throw new ConfigException("Invalid listener format: " + listener + ". Must be of format http:// [host]:[port]");
             }
         }
 
@@ -115,7 +115,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
     public synchronized Optional<HTTPServer> startHttpServer() {
         try {
             HTTPServer httpServer = new HTTPServer(listener.host, listener.port);
-            LOG.info("HTTP server started on listener " + listener);
+            LOG.info("HTTP server started on listener " + listener.host + ":" + httpServer.getPort());
             return Optional.of(httpServer);
         } catch (BindException be) {
             LOG.info("HTTP server already started");
@@ -130,10 +130,9 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
 
         @Override
         public void ensureValid(String name, Object value) {
-            String str = (String) value;
-            Matcher matcher = Listener.PATTERN.matcher(str);
+            Matcher matcher = Listener.PATTERN.matcher(String.valueOf(value));
             if (!matcher.matches()) {
-                throw new ConfigException(value + " is an invalid listener format. Must be of format http:// [host]:[port] ");
+                throw new ConfigException("Invalid listener format: " + value + ". Must be of format http:// [host]:[port]");
             }
         }
     }
--- a/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
+++ b/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
@@ -81,7 +81,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
                 int port = Integer.parseInt(matcher.group(2));
                 return new Listener(host, port);
             } else {
-                throw new ConfigException("Invalid listener format: " + listener + ". Must be of format http:// [host]:[port]");
+                throw new ConfigException(LISTENER_CONFIG + ": " +  listener +  " Listener be of format http:// [host]:[port]");
             }
         }
 
@@ -115,7 +115,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
     public synchronized Optional<HTTPServer> startHttpServer() {
         try {
             HTTPServer httpServer = new HTTPServer(listener.host, listener.port);
-            LOG.info("HTTP server started on listener " + listener.host + ":" + httpServer.getPort());
+            LOG.info("HTTP server started on listener " + "http://" +  listener.host + ":" + httpServer.getPort());
             return Optional.of(httpServer);
         } catch (BindException be) {
             LOG.info("HTTP server already started");
@@ -132,7 +132,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
         public void ensureValid(String name, Object value) {
             Matcher matcher = Listener.PATTERN.matcher(String.valueOf(value));
             if (!matcher.matches()) {
-                throw new ConfigException("Invalid listener format: " + value + ". Must be of format http:// [host]:[port]");
+                throw new ConfigException(name + ": "  + value + " Listener must be of format http:// [host]:[port]");
             }
         }
     }
--- a/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
+++ b/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
@@ -81,7 +81,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
                 int port = Integer.parseInt(matcher.group(2));
                 return new Listener(host, port);
             } else {
-                throw new ConfigException(LISTENER_CONFIG, listener, "Listener be of format http:// [host]:[port]");
+                throw new ConfigException(LISTENER_CONFIG, listener, "Listener be of format http://[host]:[port]");
             }
         }
 
@@ -132,7 +132,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
         public void ensureValid(String name, Object value) {
             Matcher matcher = Listener.PATTERN.matcher(String.valueOf(value));
             if (!matcher.matches()) {
-                throw new ConfigException(name, value, "Listener must be of format http:// [host]:[port]");
+                throw new ConfigException(name, value, "Listener must be of format http://[host]:[port]");
             }
         }
     }
--- a/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
+++ b/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
@@ -13,6 +13,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.BindException;
+import java.net.InetSocketAddress;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -21,19 +22,46 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
- * Configuration for the reporter implementations.
+ /**
+ * Configuration for the PrometheusMetricsReporter implementation.
  */
 public class PrometheusMetricsReporterConfig extends AbstractConfig {
 
     private static final Logger LOG = LoggerFactory.getLogger(PrometheusMetricsReporterConfig.class.getName());
 
+    /**
+     * Configuration prefix for Prometheus Metrics Reporter.
+     */
     public static final String CONFIG_PREFIX = "prometheus.metrics.reporter.";
+
+    /**
+     * Configuration key for the listener to expose the metrics.
+     */
     public static final String LISTENER_CONFIG = CONFIG_PREFIX + "listener";
+
+    /**
+     * Default value for the listener configuration.
+     */
     public static final String LISTENER_CONFIG_DEFAULT = "http://:8080";
+
+    /**
+     * Documentation for the listener configuration.
+     */
     public static final String LISTENER_CONFIG_DOC = "The HTTP listener to expose the metrics.";
 
+    /**
+     * Configuration key for the allowlist of metrics to collect.
+     */
     public static final String ALLOWLIST_CONFIG = CONFIG_PREFIX + "allowlist";
+
+    /**
+     * Default value for the allowlist configuration.
+     */
     public static final String ALLOWLIST_CONFIG_DEFAULT = ".*";
+
+    /**
+     * Documentation for the allowlist configuration.
+     */
     public static final String ALLOWLIST_CONFIG_DOC = "A comma separated list of regex Patterns to specify the metrics to collect.";
 
     private static final ConfigDef CONFIG_DEF = new ConfigDef()
@@ -43,12 +71,23 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
     private final Listener listener;
     private final Pattern allowlist;
 
+    /**
+     * Constructor.
+     *
+     * @param props the configuration properties.
+     */
     public PrometheusMetricsReporterConfig(Map<?, ?> props) {
         super(CONFIG_DEF, props);
         this.listener = Listener.parseListener(getString(LISTENER_CONFIG));
         this.allowlist = compileAllowlist(getList(ALLOWLIST_CONFIG));
     }
 
+    /**
+     * Check if a metric is allowed.
+     *
+     * @param name the name of the metric.
+     * @return true if the metric is allowed, false otherwise.
+     */
     public boolean isAllowed(String name) {
         return allowlist.matcher(name).matches();
     }
@@ -58,8 +97,52 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
         return Pattern.compile(joined);
     }
 
+    /**
+     * Gets the listener.
+     *
+     * @return the listener.
+     */
     public String listener() {
         return listener.toString();
+    }
+
+    @Override
+     public String toString() {
+        return "PrometheusMetricsReporterConfig{" +
+                "allowlist=" + allowlist +
+                ", listener=" + listener +
+                '}';
+    }
+
+    /**
+     * Start the HTTP server for exposing metrics.
+     *
+     * @return An optional HTTPServer instance if started successfully, otherwise empty.
+     */
+    public synchronized Optional<HTTPServer> startHttpServer() {
+        try {
+            InetSocketAddress address = listener.getInetSocketAddress();
+            HTTPServer httpServer = new HTTPServer(address.getPort(), true);
+            LOG.info("HTTP server started on listener http://{}:{}", listener.host, httpServer.getPort());
+            return Optional.of(httpServer);
+        } catch (BindException be) {
+            LOG.info("HTTP server already started");
+            return Optional.empty();
+        } catch (IOException ioe) {
+            LOG.error("Failed starting HTTP server", ioe);
+            throw new RuntimeException(ioe);
+        }
+    }
+
+    static class ListenerValidator implements ConfigDef.Validator {
+
+        @Override
+         public void ensureValid(String name, Object value) {
+            Matcher matcher = Listener.PATTERN.matcher(String.valueOf(value));
+            if (!matcher.matches()) {
+                throw new ConfigException(name, value, "Listener must be of format http://[host]:[port]");
+            }
+        }
     }
 
     static class Listener {
@@ -69,12 +152,14 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
         final String host;
         final int port;
 
-        public Listener(String host, int port) {
+        Listener(String host, int port) {
+
+
             this.host = host;
             this.port = port;
         }
 
-        public static Listener parseListener(String listener) {
+        static Listener parseListener(String listener) {
             Matcher matcher = PATTERN.matcher(listener);
             if (matcher.matches()) {
                 String host = matcher.group(1);
@@ -83,6 +168,10 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
             } else {
                 throw new ConfigException(LISTENER_CONFIG, listener, "Listener be of format http://[host]:[port]");
             }
+        }
+
+        public InetSocketAddress getInetSocketAddress() {
+            return host == null ? new InetSocketAddress(port) : InetSocketAddress.createUnresolved(host, port);
         }
 
         @Override
@@ -101,39 +190,6 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
         @Override
          public int hashCode() {
             return Objects.hash(host, port);
-        }
-    }
-
-    @Override
-    public String toString() {
-        return "PrometheusMetricsReporterConfig{" +
-                "allowlist=" + allowlist +
-                ", listener=" + listener +
-                '}';
-    }
-
-    public synchronized Optional<HTTPServer> startHttpServer() {
-        try {
-            HTTPServer httpServer = new HTTPServer(listener.host, listener.port, true);
-            LOG.info("HTTP server started on listener " + "http://" +  listener.host + ":" + httpServer.getPort());
-            return Optional.of(httpServer);
-        } catch (BindException be) {
-            LOG.info("HTTP server already started");
-            return Optional.empty();
-        } catch (IOException ioe) {
-            LOG.error("Failed starting HTTP server", ioe);
-            throw new RuntimeException(ioe);
-        }
-    }
-
-    static class ListenerValidator implements ConfigDef.Validator {
-
-        @Override
-        public void ensureValid(String name, Object value) {
-            Matcher matcher = Listener.PATTERN.matcher(String.valueOf(value));
-            if (!matcher.matches()) {
-                throw new ConfigException(name, value, "Listener must be of format http://[host]:[port]");
-            }
         }
     }
 }
--- a/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
+++ b/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
@@ -13,7 +13,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.BindException;
-import java.net.InetSocketAddress;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -22,17 +21,12 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
- /**
- * Configuration for the PrometheusMetricsReporter implementation.
- */
+* Configuration for the PrometheusMetricsReporter implementation.
+*/
 public class PrometheusMetricsReporterConfig extends AbstractConfig {
 
     private static final Logger LOG = LoggerFactory.getLogger(PrometheusMetricsReporterConfig.class.getName());
-
-    /**
-     * Configuration prefix for Prometheus Metrics Reporter.
-     */
-    public static final String CONFIG_PREFIX = "prometheus.metrics.reporter.";
+    private static final String CONFIG_PREFIX = "prometheus.metrics.reporter.";
 
     /**
      * Configuration key for the listener to expose the metrics.
@@ -43,7 +37,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
      * Default value for the listener configuration.
      */
     public static final String LISTENER_CONFIG_DEFAULT = "http://:8080";
-    private static final String LISTENER_CONFIG_DOC = "The HTTP listener to expose the metrics.";
+    public static final String LISTENER_CONFIG_DOC = "The HTTP listener to expose the metrics.";
 
     /**
      * Configuration key for the allowlist of metrics to collect.
@@ -54,7 +48,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
      * Default value for the allowlist configuration.
      */
     public static final String ALLOWLIST_CONFIG_DEFAULT = ".*";
-    private static final String ALLOWLIST_CONFIG_DOC = "A comma separated list of regex Patterns to specify the metrics to collect.";
+    private static final String ALLOWLIST_CONFIG_DOC = "A comma separated list of regex patterns to specify the metrics to collect.";
 
     private static final ConfigDef CONFIG_DEF = new ConfigDef()
             .define(LISTENER_CONFIG, ConfigDef.Type.STRING, LISTENER_CONFIG_DEFAULT, new ListenerValidator(), ConfigDef.Importance.HIGH, LISTENER_CONFIG_DOC)
@@ -99,7 +93,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
     }
 
     @Override
-     public String toString() {
+    public String toString() {
         return "PrometheusMetricsReporterConfig{" +
                 "allowlist=" + allowlist +
                 ", listener=" + listener +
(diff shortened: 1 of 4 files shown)


mimaison (commented):
Thanks for the PR! I left a few suggestions/questions

OwenCorrigan76:
@mimaison I implemented the changes around using native HttpServer.class methods.

mimaison:
Can you resolve the conflict on `src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java`?

Also the PR is still marked as draft, if it's ready for review, change its status. Thanks

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -64,22 +63,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+
+        public String getHost() {
+            return host;
+        }
+
+        public int getPort() {
+            return port;
+        }
+
+        public static HostPort parseListener(String listener) throws IOException {
```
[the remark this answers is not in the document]
OwenCorrigan76: Added

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -64,14 +63,57 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+            if (listener.equals("http://"))
+                throw new IOException("Invalid listener: Must contain a port number");
+            if (listener.startsWith("http://")) {
+                char targetLastColon = ':';
+                int lastIndex = listener.lastIndexOf(targetLastColon);
+                String host = listener.substring("http://".length(), lastIndex);
+                System.out.println("Host: " + host);
+                if (host.isEmpty()) {
+                    host = "localhost";
```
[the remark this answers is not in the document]
OwenCorrigan76: Changed

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+        final int port;
+
+        public Listener(String host, int port) {
+            this.host = host;
+            this.port = port;
+        }
+
+        public static Listener parseListener(String listener) {
+            Matcher matcher = PATTERN.matcher(listener);
+            if (matcher.find()) {
```
mimaison, line 79: Since the Validator should have checked this, we could remove this `if`. Otherwise we should use `matches()` instead of `find()`

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+                ", listener=" + listener +
                 '}';
     }
 
     public synchronized Optional<HTTPServer> startHttpServer() {
         try {
-            HTTPServer httpServer = new HTTPServer(port, true);
-            LOG.info("HTTP server started on port " + port);
+            HTTPServer httpServer = new HTTPServer(listener.host, listener.port);
+            LOG.info("HTTP server started on listener " + listener);
```
mimaison, line 118: This is not really helpful when you set the port to 0 as it won't show the actual port that it picked.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+        }
+
+        public static Listener parseListener(String listener) {
+            Matcher matcher = PATTERN.matcher(listener);
+            if (matcher.find()) {
+                String host = matcher.group(1);
+                int port = Integer.parseInt(matcher.group(2));
+                return new Listener(host, port);
+            } else {
+                throw new ConfigException("Invalid listener format: " + listener);
```
mimaison, line 84: Can we use the constructor with 3 arguments?
https://kafka.apache.org/36/javadoc/org/apache/kafka/common/config/ConfigException.html#%3Cinit%3E(java.lang.String,java.lang.Object,java.lang.String)

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -82,4 +126,15 @@ public synchronized Optional<HTTPServer> startHttpServer() {
...
     }
 
+    static class ListenerValidator implements ConfigDef.Validator {
+
+        @Override
+        public void ensureValid(String name, Object value) {
+            String str = (String) value;
+            Matcher matcher = Listener.PATTERN.matcher(str);
+            if (!matcher.matches()) {
+                throw new ConfigException(value + " is an invalid listener format. Must be of format http:// [host]:[port] ");
```
mimaison, line 136: Can we use the constructor with 3 arguments?
https://kafka.apache.org/36/javadoc/org/apache/kafka/common/config/ConfigException.html#%3Cinit%3E(java.lang.String,java.lang.Object,java.lang.String)

src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java
```diff
@@ -45,4 +47,20 @@ public void testAllowlist() {
...
         assertTrue(config.isAllowed("kafka_network_metric"));
     }
+
+    @Test
+    public void testListenerParseListener() {
+        assertEquals(new PrometheusMetricsReporterConfig.Listener("", 8080), PrometheusMetricsReporterConfig.Listener.parseListener("http://:8080"));
+        assertEquals(new PrometheusMetricsReporterConfig.Listener("123", 8080), PrometheusMetricsReporterConfig.Listener.parseListener("http://123:8080"));
+        assertEquals(new PrometheusMetricsReporterConfig.Listener("::1", 8080), PrometheusMetricsReporterConfig.Listener.parseListener("http://::1:8080"));
+        assertEquals(new PrometheusMetricsReporterConfig.Listener("::1", 8080), PrometheusMetricsReporterConfig.Listener.parseListener("http://[::1]:8080"));
+        assertEquals(new PrometheusMetricsReporterConfig.Listener("random", 8080), PrometheusMetricsReporterConfig.Listener.parseListener("http://random:8080"));
```
mimaison, line 57: Probably worth adding a test like:
`assertThrows(ConfigException.class, () -> PrometheusMetricsReporterConfig.Listener.parseListener("abchttp://:8080asd"));`

This would currently fail because we're using `matcher.find()` instead of matcher.matches() in `parseListener()`

src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java
```diff
@@ -45,4 +47,20 @@ public void testAllowlist() {
...
+
+        assertThrows(ConfigException.class, () -> PrometheusMetricsReporterConfig.Listener.parseListener("http"));
+        assertThrows(ConfigException.class, () -> PrometheusMetricsReporterConfig.Listener.parseListener("http://"));
+        assertThrows(ConfigException.class, () -> PrometheusMetricsReporterConfig.Listener.parseListener("http://random"));
+        assertThrows(ConfigException.class, () -> PrometheusMetricsReporterConfig.Listener.parseListener("http://random:"));
+        assertThrows(ConfigException.class, () -> PrometheusMetricsReporterConfig.Listener.parseListener("http://:-8080"));
+        assertThrows(ConfigException.class, () -> PrometheusMetricsReporterConfig.Listener.parseListener("http://random:-8080"));
+    }
```
mimaison, line 65: Can we add a test for the Validator?

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -82,4 +126,15 @@ public synchronized Optional<HTTPServer> startHttpServer() {
         }
     }
 
+    static class ListenerValidator implements ConfigDef.Validator {
+
+        @Override
+        public void ensureValid(String name, Object value) {
+            String str = (String) value;
```
mimaison, line 133: We can inline this variable

mimaison (commented):
Thanks for the PR. It looks good overall, I left a few suggestions/questions.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Changed to `matches`

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Changed to `LOG.info("HTTP server started on listener " + listener.host + ":" + httpServer.getPort());`

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Do you mean like this?
```throw new ConfigException("Invalid listener format: " + value + ". Must be of format http:// [host]:[port]");```

File: src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java (hunk 6 above)
Comment (OwenCorrigan76) [reply src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java]: Do you mean like this?:
```throw new ConfigException("Invalid listener format: " + value + ". Must be of format http:// [host]:[port]");```

File: src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java (hunk 7 above)
Comment (OwenCorrigan76) [reply src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java:57]: Added

File: src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java (hunk 8 above)
Comment (OwenCorrigan76) [reply src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java:80]: Is what I added sufficient?

File: src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java (hunk 9 above)
Comment (OwenCorrigan76) [reply src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java]: Done

File: src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java (hunk 6 above)
Comment (mimaison) [reply src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java]: I mean something like `throw new ConfigException(name, value, "Listener must be of format http:// [host]:[port]");`
Same above

File: src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java (hunk 4 above)
Comment (mimaison) [reply src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java]: Can we also add `http://` before the host?

File: src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java (hunk 7 above)
Comment (mimaison) [reply src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java:57]: To be exhaustive we can also try with `randomhttp://:8080` and `http://:8080random`

Comment (OwenCorrigan76) [reply src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java:57]: Added

File: src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java (hunk 6 above)
Comment (OwenCorrigan76) [reply src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java]: Done

Comment (mimaison) [reply src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java]: You don't seem to have addressed this. The code is still using the constructor with a single argument.

File: src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java (hunk 8 above)
Comment (mimaison) [reply src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java:82]: I think we could use the same test cases than in `testListenerParseListener()`

Comment (OwenCorrigan76) [reply src/test/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfigTest.java:88]: Done

File: src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java [hunk 10]
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+        }
+
+        public static Listener parseListener(String listener) {
+            Matcher matcher = PATTERN.matcher(listener);
+            if (matcher.matches()) {
+                String host = matcher.group(1);
+                int port = Integer.parseInt(matcher.group(2));
+                return new Listener(host, port);
+            } else {
+                throw new ConfigException(LISTENER_CONFIG, listener, "Listener be of format http:// [host]:[port]");
```
mimaison, line 84: Can we remove the space after `http://`? Same below

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
-                ", port=" + port +
+                ", listener=" + listener +
                 '}';
     }
 
     public synchronized Optional<HTTPServer> startHttpServer() {
         try {
-            HTTPServer httpServer = new HTTPServer(port, true);
-            LOG.info("HTTP server started on port " + port);
+            HTTPServer httpServer = new HTTPServer(listener.host, listener.port);
```
mimaison, line 117: We still need to set `daemon` to `true`.

OwenCorrigan76: Fixed

mimaison (approved):
LGTM, thanks for the PR

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
         return Pattern.compile(joined);
     }
 
+    public String listener() {
+        return listener.toString();
+    }
+
+    static class Listener {
```
scholzj, line 65: Would it make sense to for example put all the inner classes at the end of the parent class?

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+        public static Listener parseListener(String listener) {
+            Matcher matcher = PATTERN.matcher(listener);
+            if (matcher.matches()) {
+                String host = matcher.group(1);
+                int port = Integer.parseInt(matcher.group(2));
+                return new Listener(host, port);
+            } else {
+                throw new ConfigException(LISTENER_CONFIG, listener, "Listener be of format http://[host]:[port]");
+            }
+        }
```
scholzj, line 86: Do these methods need to be public?

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
scholzj, line 1: Please add Javadocs to the public (non-override) methods and fields you added.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+        final String host;
+        final int port;
+
+        public Listener(String host, int port) {
+            this.host = host;
+            this.port = port;
+        }
+
+        public static Listener parseListener(String listener) {
+            Matcher matcher = PATTERN.matcher(listener);
```
k-wall, line 78: Why not just lean on java.net.URL() to do the parsing for you?

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+                ", listener=" + listener +
                 '}';
     }
 
     public synchronized Optional<HTTPServer> startHttpServer() {
         try {
-            HTTPServer httpServer = new HTTPServer(port, true);
-            LOG.info("HTTP server started on port " + port);
+            HTTPServer httpServer = new HTTPServer(listener.host, listener.port, true);
+            LOG.info("HTTP server started on listener " + "http://" +  listener.host + ":" + httpServer.getPort());
```
k-wall, line 118: Use replacement patterns rather than interpolation to build the string to log.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
mimaison: I briefly looked into this and found that it was harder to get what we wanted. It did not seem to support having an empty host and then you had to check that there were no path, query parameters, fragments, etc

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+    }
+
+    static class Listener {
+
+        private static final Pattern PATTERN = Pattern.compile("http://\\[?([0-9a-zA-Z\\-%._:]*)]?:([0-9]+)");
+
+        final String host;
+        final int port;
+
+        public Listener(String host, int port) {
```
k-wall, line 72: you are using a parseListener as a factory method for Listener objects (good), so that means you can make the ctor private.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
mimaison: The regex used is "inspired" from the regex used in Kafka Connect to parse the listener config: https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/rest/RestServer.java#L83

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Public removed. Now package private

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Added. I'm not 100% sure if the're ok

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +58,64 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
-                ", port=" + port +
+                ", listener=" + listener +
                 '}';
     }
 
     public synchronized Optional<HTTPServer> startHttpServer() {
         try {
-            HTTPServer httpServer = new HTTPServer(port, true);
-            LOG.info("HTTP server started on port " + port);
+            HTTPServer httpServer = new HTTPServer(listener.host, listener.port, true);
```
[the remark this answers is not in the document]
OwenCorrigan76: I'm still not sure if I have implemented this correctly?? Do you think it would be better to run `startHttpServer` in the `init` method instead of the `configure` method?

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Done

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +97,33 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+     * Start the HTTP server for exposing metrics.
+     *
+     * @return An optional HTTPServer instance if started successfully, otherwise empty.
+     */
     public synchronized Optional<HTTPServer> startHttpServer() {
         try {
-            HTTPServer httpServer = new HTTPServer(port, true);
-            LOG.info("HTTP server started on port " + port);
+            InetSocketAddress address = listener.getInetSocketAddress();
+            HTTPServer httpServer = new HTTPServer(address.getPort(), true);
```
k-wall, line 125:
```suggestion
                        HTTPServer httpServer = new HTTPServer(address, CollectorRegistry.defaultRegistry, true);

```

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -82,4 +134,62 @@ public synchronized Optional<HTTPServer> startHttpServer() {
...
+                String host = matcher.group(1);
+                int port = Integer.parseInt(matcher.group(2));
+                return new Listener(host, port);
+            } else {
+                throw new ConfigException(LISTENER_CONFIG, listener, "Listener be of format http://[host]:[port]");
+            }
+        }
+
+        public InetSocketAddress getInetSocketAddress() {
+            return host == null ? new InetSocketAddress(port) : InetSocketAddress.createUnresolved(host, port);
```
k-wall, line 174:
```suggestion
            return host == null || host.isEmpty() ? new InetSocketAddress(port) : InetSocketAddress.createUnresolved(host, port);
```

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -82,4 +134,62 @@ public synchronized Optional<HTTPServer> startHttpServer() {
...
+    }
+
+    static class Listener {
+
+        private static final Pattern PATTERN = Pattern.compile("http://\\[?([0-9a-zA-Z\\-%._:]*)]?:([0-9]+)");
+
+        final String host;
+        final int port;
+
+        Listener(String host, int port) {
```
k-wall, line 155: I wonder if you actually need the Listener class at all.  You could just have parseListener return a InetSocketAddress....

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -7,50 +7,87 @@
...
+ * Configuration for the PrometheusMetricsReporter implementation.
  */
 public class PrometheusMetricsReporterConfig extends AbstractConfig {
 
     private static final Logger LOG = LoggerFactory.getLogger(PrometheusMetricsReporterConfig.class.getName());
 
+    /**
+     * Configuration prefix for Prometheus Metrics Reporter.
+     */
     public static final String CONFIG_PREFIX = "prometheus.metrics.reporter.";
```
mimaison, line 35: This field could be private so you don't have to add javadoc

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -7,50 +7,87 @@
...
 
+    /**
+     * Default value for the listener configuration.
+     */
+    public static final String LISTENER_CONFIG_DEFAULT = "http://:8080";
+
+    /**
+     * Documentation for the listener configuration.
+     */
+    public static final String LISTENER_CONFIG_DOC = "The HTTP listener to expose the metrics.";
```
mimaison, line 50: All the `DOC` fields could be private too

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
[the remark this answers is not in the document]
mimaison: If the hostname is empty, `InetAddress.getAllByName()` which is called by the `InetSocketAddress()` constructor returns the loopback address


From https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#getAllByName-java.lang.String-
> If the host is null then an InetAddress representing an address of the loopback interface is returned.

But you're right to be concerned by doing a hostname resolution at startup, I had missed that point. Looking at the code, it seems `HttpServer.create()` ends up calling `new ServerImpl()` which calls `socket.bind()` on the `InetSocketAddress`. This throws `SocketException` if the address is unresolved. I've not tried running the code, so it's worth double checking but I think this is why I told Owen to use the `HttpServer` constructor with a host/port.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
mimaison: We considered that option, and I think it makes sense if we switch to use an `InetSocketAddress`. We decided to keep a container class as I expect we'll add https in the near future.

mimaison: We should probably have had these discussions in public so you (and others) could see our thought process.

OwenCorrigan76: Does it make sense to keep the `Listener` class in with the prospect of adding https?

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Changed to private

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Changed to private

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: @mimaison @k-wall I've added the InetSocketAddress but kept the Listener class. Everything looks to be working as expected though I am open to correction.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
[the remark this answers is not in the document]
mimaison: I wonder if we actually want name resolution at early as possible if the user provided a hostname. Until the HTTP server is started, we won't export metrics and if we can't resolve it we'd rather fail fast than the broker for a moment without exporting metrics.

For reference, this is what jmx_exporter does:
- `premain()`: https://github.com/prometheus/jmx_exporter/blob/main/jmx_prometheus_javaagent/src/main/java/io/prometheus/jmx/JavaAgent.java#L60-L66
- `createHTTPServer()`: https://github.com/prometheus/jmx_exporter/blob/main/jmx_prometheus_common/src/main/java/io/prometheus/jmx/common/http/HTTPServerFactory.java#L104-L115

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -58,11 +54,7 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
      * Default value for the allowlist configuration.
      */
     public static final String ALLOWLIST_CONFIG_DEFAULT = ".*";
-
-    /**
-     * Documentation for the allowlist configuration.
-     */
-    public static final String ALLOWLIST_CONFIG_DOC = "A comma separated list of regex Patterns to specify the metrics to collect.";
+    private static final String ALLOWLIST_CONFIG_DOC = "A comma separated list of regex Patterns to specify the metrics to collect.";
```
k-wall, line 57: I don't think Pattern deserves a capital.
```suggestion
    private static final String ALLOWLIST_CONFIG_DOC = "A comma separated list of regex patterns that specify the metrics to collect.";
```

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
k-wall: It is still public

k-wall (approved):

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
mimaison: Considering the discussion above, I think we should just call `new HTTPServer(host, port, true)`

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -82,4 +126,60 @@ public synchronized Optional<HTTPServer> startHttpServer() {
...
+                String host = matcher.group(1);
+                int port = Integer.parseInt(matcher.group(2));
+                return new Listener(host, port);
+            } else {
+                throw new ConfigException(LISTENER_CONFIG, listener, "Listener be of format http://[host]:[port]");
+            }
+        }
+
+        public InetSocketAddress getInetSocketAddress() {
+            return host == null || host.isEmpty() ? new InetSocketAddress(port) : InetSocketAddress.createUnresolved(host, port);
```
mimaison, line 164: If we do https://github.com/strimzi/metrics-reporter/pull/16/files#r1504056500 then we can delete this method

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
k-wall: Ok

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -7,50 +7,79 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
- * Configuration for the reporter implementations.
+ /**
```
scholzj, line 25: Do you have twice `/**`? Or is that some bad rendering in GitHub UI?

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -60,18 +89,33 @@ private Pattern compileAllowlist(List<String> allowlist) {
...
+     *
+     * @return the listener.
+     */
+    public String listener() {
+        return listener.toString();
+    }
+
     @Override
-    public String toString() {
+     public String toString() {
```
scholzj, line 102:
```suggestion
    public String toString() {
```

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -82,4 +126,60 @@ public synchronized Optional<HTTPServer> startHttpServer() {
...
+            if (matcher.matches()) {
+                String host = matcher.group(1);
+                int port = Integer.parseInt(matcher.group(2));
+                return new Listener(host, port);
+            } else {
+                throw new ConfigException(LISTENER_CONFIG, listener, "Listener be of format http://[host]:[port]");
+            }
+        }
+
+        public InetSocketAddress getInetSocketAddress() {
```
scholzj, line 163: Should this be public in non-public class? If yes, please add Javadoc.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
mimaison: This field is still `public`

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Both inner classes moved to the end of the parent class

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: CONFIG_PREFIX now private

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Capital P changed to lower case

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Method removed

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Duplicate removed

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: Moved back

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
OwenCorrigan76: This is now not being used

scholzj (approved):
LGTM.

k-wall (approved):

mimaison (approved):
LGTM, thanks for the PR!

Issue #8, opened 2024-01-30
Handle non-numeric metric values


mimaison:
As per [the proposal](https://github.com/strimzi/proposals/blob/main/064-prometheus-metrics-reporter.md), non-numeric values should be handled the following way:

> Out of all the metrics emitted by Kafka brokers and clients, a few of them have non-numeric values. Prometheus only supports numeric values for metrics. When using jmx_exporter it is possible to write rules that move the non-numeric value to a label. For example we do it in [kafka-connect-metrics](https://github.com/strimzi/strimzi-kafka-operator/blob/main/packaging/examples/metrics/kafka-connect-metrics.yaml#L103-L111) for the status metric which has a string value.
>
> I propose to do this automatically in the reporters. For example a metric named: kafka.connect:type=connector-task-metrics,connector="{connector}",task="{task}"<>status with the value running will be converted into kafka_connect_connector_task_metrics_status{connector="heartbeats",task="0",status="running"} and its value will be set to 1.0.

mimaison:
Done in https://github.com/strimzi/metrics-reporter/pull/21

Issue #6, opened 2024-01-30
Cleanup logging


mimaison:
At the moment there is a lot of logging for debugging purposes. We should clean that up and keep important debugging logs with trace or debug levels.

mimaison:
The point of this ticket is to review the current logging and adjust it so it's relevant and not too verbose, especially at INFO level which is usually the default.

I can't comment on each example. For example, while you were working on the reporter which logs did you find useful?

mimaison:
Fixed in https://github.com/strimzi/metrics-reporter/pull/33

PR #21, opened 2024-06-10
Handle non-numeric metric values

OwenCorrigan76:
Prometheus only supports numeric values for metrics. A few metrics emitted by Kafka brokers and clients have non-numeric values. This PR will move the non-numeric value to a label and then and set its value to 1.0,  to satisfy Prometheus only supporting numeric values.

Reported in the following issue:
https://github.com/strimzi/metrics-reporter/issues/8

### DIFF
--- a/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
@@ -111,12 +111,6 @@ public class KafkaMetricsCollector extends Collector {
 
     static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels) {
         Object value = metric.metricValue();
-        if (!(value instanceof Number)) {
-            // Prometheus only accepts numeric metrics.
-            // Kafka gauges can have arbitrary types, so skip them for now
-            // TODO move non-numeric values to labels
-            return null;
-        }
         Map<String, String> sanitizedLabels = labels.entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
@@ -125,6 +119,16 @@ public class KafkaMetricsCollector extends Collector {
                             throw new IllegalStateException("Unexpected duplicate key " + v1);
                         },
                         LinkedHashMap::new));
+
+        if (!(value instanceof Number)) {
+            // If the value is non-numeric, add it as a label and set the metric value to 1.0
+            LOG.info("*******Converting non-numeric metric {} with value {} to label", name, value);
+            sanitizedLabels.put("value", value.toString());
+            return new MetricFamilySamplesBuilder(Type.GAUGE, help)
+                    .addSample(name, 1.0, sanitizedLabels)
+                    .build();
+        }
+
         return new MetricFamilySamplesBuilder(Type.GAUGE, help)
                 .addSample(name, ((Number) value).doubleValue(), sanitizedLabels)
                 .build();
--- a/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
@@ -110,7 +110,8 @@ public class KafkaMetricsCollector extends Collector {
     }
 
     static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels) {
-        Object value = metric.metricValue();
+        Object valueObj = metric.metricValue();
+        final double value;
         Map<String, String> sanitizedLabels = labels.entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
@@ -120,17 +121,16 @@ public class KafkaMetricsCollector extends Collector {
                         },
                         LinkedHashMap::new));
 
-        if (!(value instanceof Number)) {
-            // If the value is non-numeric, add it as a label and set the metric value to 1.0
-            LOG.info("*******Converting non-numeric metric {} with value {} to label", name, value);
-            sanitizedLabels.put("value", value.toString());
-            return new MetricFamilySamplesBuilder(Type.GAUGE, help)
-                    .addSample(name, 1.0, sanitizedLabels)
-                    .build();
+        if (valueObj instanceof Number) {
+            value = ((Number) valueObj).doubleValue();
+        } else {
+            value = 1.0;
+            sanitizedLabels.put(name, String.valueOf(valueObj));
         }
 
         return new MetricFamilySamplesBuilder(Type.GAUGE, help)
-                .addSample(name, ((Number) value).doubleValue(), sanitizedLabels)
+                .addSample(name, value, sanitizedLabels)
                 .build();
     }
+
 }
--- a/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
@@ -59,14 +59,14 @@ public class KafkaMetricsCollector extends Collector {
             KafkaMetric kafkaMetric = entry.getValue();
             LOG.trace("Collecting Kafka metric {}", metricName);
 
-            String name = metricName(metricName);
+            String prometheusMetricName = metricName(metricName);
             // TODO Filtering should take labels into account
-            if (!config.isAllowed(name)) {
-                LOG.info("Kafka metric {} is not allowed", name);
+            if (!config.isAllowed(prometheusMetricName)) {
+                LOG.info("Kafka metric {} is not allowed", prometheusMetricName);
                 continue;
             }
-            LOG.info("Kafka metric {} is allowed", name);
-            MetricFamilySamples sample = convert(name, metricName.description(), kafkaMetric, metricName.tags(), metricName);
+            LOG.info("Kafka metric {} is allowed", prometheusMetricName);
+            MetricFamilySamples sample = convert(prometheusMetricName, kafkaMetric, metricName);
             if (sample != null) {
                 samples.add(sample);
             }
@@ -108,13 +108,13 @@ public class KafkaMetricsCollector extends Collector {
         return prefix + '_' + group + '_' + name;
     }
 
-    static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels, MetricName metricName) {
-        Map<String, String> sanitizedLabels = labels.entrySet().stream()
+    private static MetricFamilySamples convert(String prometheusMetricName, KafkaMetric metric, MetricName metricName) {
+        Map<String, String> sanitizedLabels = metricName.tags().entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
                         Map.Entry::getValue,
                         (v1, v2) -> {
-                            LOG.warn("Unexpected duplicate key" +  v1);
+                            LOG.warn("Ignoring metric value duplicate key {}", v1);
                             return v1;
                         },
                         LinkedHashMap::new));
@@ -128,8 +128,8 @@ public class KafkaMetricsCollector extends Collector {
             sanitizedLabels.put(attributeName, String.valueOf(valueObj));
         }
 
-        return new MetricFamilySamplesBuilder(Type.GAUGE, help)
-                .addSample(name, value, sanitizedLabels)
+        return new MetricFamilySamplesBuilder(Type.GAUGE, metric.metricName().description())
+                .addSample(prometheusMetricName, value, sanitizedLabels)
                 .build();
     }
 }
--- a/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
@@ -111,7 +111,7 @@ public class KafkaMetricsCollector extends Collector {
 
     static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels) {
         Object valueObj = metric.metricValue();
-        final double value;
+        double value;
         Map<String, String> sanitizedLabels = labels.entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
--- a/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
@@ -66,8 +66,7 @@ public class KafkaMetricsCollector extends Collector {
                 continue;
             }
             LOG.info("Kafka metric {} is allowed", name);
-            LOG.info("labels " + metricName.tags());
-            MetricFamilySamples sample = convert(name, metricName.description(), kafkaMetric, metricName.tags());
+            MetricFamilySamples sample = convert(name, metricName.description(), kafkaMetric, metricName.tags(), metricName);
             if (sample != null) {
                 samples.add(sample);
             }
@@ -109,28 +108,28 @@ public class KafkaMetricsCollector extends Collector {
         return prefix + '_' + group + '_' + name;
     }
 
-    static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels) {
-        Object valueObj = metric.metricValue();
-        double value;
+    static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels, MetricName metricName) {
         Map<String, String> sanitizedLabels = labels.entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
                         Map.Entry::getValue,
                         (v1, v2) -> {
-                            throw new IllegalStateException("Unexpected duplicate key " + v1);
+                            LOG.warn("Unexpected duplicate key" +  v1);
+                            return v1;
                         },
                         LinkedHashMap::new));
-
+        Object valueObj = metric.metricValue();
+        double value;
         if (valueObj instanceof Number) {
             value = ((Number) valueObj).doubleValue();
         } else {
             value = 1.0;
-            sanitizedLabels.put(name, String.valueOf(valueObj));
+            String attributeName = metricName.name();
+            sanitizedLabels.put(attributeName, String.valueOf(valueObj));
         }
 
         return new MetricFamilySamplesBuilder(Type.GAUGE, help)
                 .addSample(name, value, sanitizedLabels)
                 .build();
     }
-
 }
--- a/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
@@ -12,12 +12,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 
 /**
  * Prometheus Collector to store and export metrics retrieved by the reporters.
@@ -109,15 +107,7 @@ public class KafkaMetricsCollector extends Collector {
     }
 
     private static MetricFamilySamples convert(String prometheusMetricName, KafkaMetric metric, MetricName metricName) {
-        Map<String, String> sanitizedLabels = metricName.tags().entrySet().stream()
-                .collect(Collectors.toMap(
-                        e -> Collector.sanitizeMetricName(e.getKey()),
-                        Map.Entry::getValue,
-                        (v1, v2) -> {
-                            LOG.warn("Ignoring metric value duplicate key {}", v1);
-                            return v1;
-                        },
-                        LinkedHashMap::new));
+        Map<String, String> sanitizedLabels = MetricFamilySamplesBuilder.sanitizeLabels(metricName.tags());
         Object valueObj = metric.metricValue();
         double value;
         if (valueObj instanceof Number) {
@@ -125,11 +115,11 @@ public class KafkaMetricsCollector extends Collector {
         } else {
             value = 1.0;
             String attributeName = metricName.name();
-            sanitizedLabels.put(attributeName, String.valueOf(valueObj));
+            sanitizedLabels.put(Collector.sanitizeMetricName(attributeName), String.valueOf(valueObj));
         }
 
         return new MetricFamilySamplesBuilder(Type.GAUGE, metric.metricName().description())
-                .addSample(prometheusMetricName, value, sanitizedLabels)
-                .build();
+                   .addSample(prometheusMetricName, value, sanitizedLabels)
+                   .build();
     }
 }
--- a/src/main/java/io/strimzi/kafka/metrics/MetricFamilySamplesBuilder.java
+++ b/src/main/java/io/strimzi/kafka/metrics/MetricFamilySamplesBuilder.java
@@ -64,6 +64,15 @@ public class MetricFamilySamplesBuilder {
         return new Collector.MetricFamilySamples(samples.get(0).name, type, help, samples);
     }
 
+    /**
+     * Sanitizes the given map of labels by replacing any characters in the label keys
+     * that are not allowed in Prometheus metric names with an underscore ('_').
+     * If there are duplicate keys after sanitization, a warning is logged, and the first value is retained.
+     *
+     * @param labels The map of labels to be sanitized. The keys of this map are label names,
+     *               and the values are label values.
+     * @return A new map with sanitized label names and the same label values.
+     */
     public static Map<String, String> sanitizeLabels(Map<String, String> labels) {
         return labels.entrySet().stream()
                 .collect(Collectors.toMap(
--- a/src/main/java/io/strimzi/kafka/metrics/MetricFamilySamplesBuilder.java
+++ b/src/main/java/io/strimzi/kafka/metrics/MetricFamilySamplesBuilder.java
@@ -22,7 +22,7 @@ import java.util.stream.Collectors;
  */
 public class MetricFamilySamplesBuilder {
 
-    private static final Logger LOG = LoggerFactory.getLogger(KafkaMetricsCollector.class.getName());
+    private static final Logger LOG = LoggerFactory.getLogger(MetricFamilySamplesBuilder.class.getName());
     private final Collector.Type type;
     private final String help;
     private final List<Collector.MetricFamilySamples.Sample> samples;
--- a/src/main/java/io/strimzi/kafka/metrics/MetricFamilySamplesBuilder.java
+++ b/src/main/java/io/strimzi/kafka/metrics/MetricFamilySamplesBuilder.java
@@ -6,18 +6,23 @@ package io.strimzi.kafka.metrics;
 
 import com.yammer.metrics.stats.Snapshot;
 import io.prometheus.client.Collector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Helper class to convert Kafka metrics into the Prometheus format.
  */
 public class MetricFamilySamplesBuilder {
 
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaMetricsCollector.class.getName());
     private final Collector.Type type;
     private final String help;
     private final List<Collector.MetricFamilySamples.Sample> samples;
@@ -57,5 +62,17 @@ public class MetricFamilySamplesBuilder {
             throw new IllegalStateException("There are no samples");
         }
         return new Collector.MetricFamilySamples(samples.get(0).name, type, help, samples);
+    }
+
+    public static Map<String, String> sanitizeLabels(Map<String, String> labels) {
+        return labels.entrySet().stream()
+                .collect(Collectors.toMap(
+                        e -> Collector.sanitizeMetricName(e.getKey()),
+                        Map.Entry::getValue,
+                        (v1, v2) -> {
+                            LOG.warn("Ignoring metric value duplicate key {}", v1);
+                            return v1;
+                        },
+                        LinkedHashMap::new));
     }
 }
--- a/src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
@@ -73,7 +73,7 @@ public class YammerMetricsCollector extends Collector {
                 if (metric instanceof Counter) {
                     sample = convert(name, (Counter) metric, labels);
                 } else if (metric instanceof Gauge) {
-                    sample = convert(name, (Gauge<?>) metric, labels);
+                    sample = convert(name, (Gauge<?>) metric, labels, metricName);
                 } else if (metric instanceof Histogram) {
                     sample = convert(name, (Histogram) metric, labels);
                 } else if (metric instanceof Meter) {
@@ -121,23 +121,24 @@ public class YammerMetricsCollector extends Collector {
                 .build();
     }
 
-    static MetricFamilySamples convert(String name, Gauge<?> gauge, Map<String, String> labels) {
-        Object valueObj = gauge.value();
-        double value;
+    static MetricFamilySamples convert(String name, Gauge<?> gauge, Map<String, String> labels, MetricName metricName) {
         Map<String, String> sanitizedLabels = labels.entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
                         Map.Entry::getValue,
                         (v1, v2) -> {
-                            throw new IllegalStateException("Unexpected duplicate key " + v1);
+                            LOG.warn("Unexpected duplicate key " + v1);
+                            return v1;
                         },
                         LinkedHashMap::new));
-
+        Object valueObj = gauge.value();
+        double value;
         if (valueObj instanceof Number) {
             value = ((Number) valueObj).doubleValue();
         } else {
             value = 1.0;
-            sanitizedLabels.put("value", String.valueOf(valueObj));
+            String attributeName = metricName.getName();
+            sanitizedLabels.put(attributeName, String.valueOf(valueObj));
         }
 
         return new MetricFamilySamplesBuilder(Type.GAUGE, "")
--- a/src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
@@ -59,27 +59,27 @@ public class YammerMetricsCollector extends Collector {
                 Metric metric = entry.getValue();
                 LOG.trace("Collecting Yammer metric {}", metricName);
 
-                String name = metricName(metricName);
+                String prometheusMetricName = metricName(metricName);
                 // TODO Filtering should take labels into account
-                if (!config.isAllowed(name)) {
-                    LOG.info("Yammer metric {} is not allowed", name);
+                if (!config.isAllowed(prometheusMetricName)) {
+                    LOG.info("Yammer metric {} is not allowed", prometheusMetricName);
                     continue;
                 }
-                LOG.info("Yammer metric {} is allowed", name);
+                LOG.info("Yammer metric {} is allowed", prometheusMetricName);
                 Map<String, String> labels = labelsFromScope(metricName.getScope());
                 LOG.info("labels " + labels);
 
                 MetricFamilySamples sample = null;
                 if (metric instanceof Counter) {
-                    sample = convert(name, (Counter) metric, labels);
+                    sample = convert(prometheusMetricName, (Counter) metric, labels);
                 } else if (metric instanceof Gauge) {
-                    sample = convert(name, (Gauge<?>) metric, labels, metricName);
+                    sample = convert(prometheusMetricName, (Gauge<?>) metric, labels, metricName);
                 } else if (metric instanceof Histogram) {
-                    sample = convert(name, (Histogram) metric, labels);
+                    sample = convert(prometheusMetricName, (Histogram) metric, labels);
                 } else if (metric instanceof Meter) {
-                    sample = convert(name, (Meter) metric, labels);
+                    sample = convert(prometheusMetricName, (Meter) metric, labels);
                 } else if (metric instanceof Timer) {
-                    sample = convert(name, (Timer) metric, labels);
+                    sample = convert(prometheusMetricName, (Timer) metric, labels);
                 } else {
                     LOG.error("The metric " + metric.getClass().getName() + " has an unexpected type.");
                 }
@@ -115,19 +115,19 @@ public class YammerMetricsCollector extends Collector {
         return Collections.emptyMap();
     }
 
-    static MetricFamilySamples convert(String name, Counter counter, Map<String, String> labels) {
+    static MetricFamilySamples convert(String prometheusMetricName, Counter counter, Map<String, String> labels) {
         return new MetricFamilySamplesBuilder(Type.GAUGE, "")
-                .addSample(name + "_count", counter.count(), labels)
+                .addSample(prometheusMetricName + "_count", counter.count(), labels)
                 .build();
     }
 
-    static MetricFamilySamples convert(String name, Gauge<?> gauge, Map<String, String> labels, MetricName metricName) {
+    private static MetricFamilySamples convert(String prometheusMetricName, Gauge<?> gauge, Map<String, String> labels, MetricName metricName) {
         Map<String, String> sanitizedLabels = labels.entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
                         Map.Entry::getValue,
                         (v1, v2) -> {
-                            LOG.warn("Unexpected duplicate key " + v1);
+                            LOG.warn("Ignoring metric value duplicate key {}", v1);
                             return v1;
                         },
                         LinkedHashMap::new));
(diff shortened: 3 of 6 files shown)


src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -125,6 +119,16 @@ static MetricFamilySamples convert(String name, String help, KafkaMetric metric,
                             throw new IllegalStateException("Unexpected duplicate key " + v1);
                         },
                         LinkedHashMap::new));
+
+        if (!(value instanceof Number)) {
+            // If the value is non-numeric, add it as a label and set the metric value to 1.0
+            LOG.info("*******Converting non-numeric metric {} with value {} to label", name, value);
+            sanitizedLabels.put("value", value.toString());
+            return new MetricFamilySamplesBuilder(Type.GAUGE, help)
```
k-wall, line 127: You've duplicated the code need to create the builder, violating DRY.  That creates a maintenance overhead you want to avoid.  I'd suggest this pattern:

final double value;
if (value instanceof Number) {
   value = ((Number) value).doubleValue()
} else {
   value = 1.0;
   sanitizedLabels.put("value", value.toString());
}

// create builder.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -125,6 +119,16 @@ static MetricFamilySamples convert(String name, String help, KafkaMetric metric,
                             throw new IllegalStateException("Unexpected duplicate key " + v1);
                         },
                         LinkedHashMap::new));
+
+        if (!(value instanceof Number)) {
+            // If the value is non-numeric, add it as a label and set the metric value to 1.0
+            LOG.info("*******Converting non-numeric metric {} with value {} to label", name, value);
+            sanitizedLabels.put("value", value.toString());
```
k-wall, line 126: can value ever be null?   To be defensive use String.valueOf(value) as that will never NPE on you.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -51,10 +50,15 @@ public void testCollect() {
         metrics = collector.collect();
         assertTrue(metrics.isEmpty());
 
-        // Adding a non-numeric metric does nothing
-        collector.addMetric(buildNonNumericMetric("name2", "group"));
+        // Adding a non-numeric metric converted
```
k-wall, line 53: Tests are best when they are focused on testing a single responsibility.     That makes it really clean what features are covered.    It is also more obvious what's broken when tests fail.

So if I were adding this feature, I'd be adding a new test case.. testNonNumericMetric

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -51,10 +50,15 @@ public void testCollect() {
...
+        KafkaMetric nonNumericMetric = buildNonNumericMetric("name", "group");
+        collector.addMetric(nonNumericMetric);
         metrics = collector.collect();
-        assertTrue(metrics.isEmpty());
+        assertEquals(1, metrics.size());
+        assertEquals("kafka_server_group_name", metrics.get(0).name);
+        assertEquals(1, metrics.get(0).samples.size());
+        assertEquals(1.0, metrics.get(0).samples.get(0).value, 1.0);
```
k-wall, line 60: Use 1.0 for the delta/epsilon looks odd.   That's a floating point tolerance value.  0 should work in this use-case because there is no floating point maths involved.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -51,10 +50,15 @@ public void testCollect() {
...
-        // Adding a non-numeric metric does nothing
-        collector.addMetric(buildNonNumericMetric("name2", "group"));
+        // Adding a non-numeric metric converted
+        KafkaMetric nonNumericMetric = buildNonNumericMetric("name", "group");
+        collector.addMetric(nonNumericMetric);
         metrics = collector.collect();
-        assertTrue(metrics.isEmpty());
+        assertEquals(1, metrics.size());
+        assertEquals("kafka_server_group_name", metrics.get(0).name);
+        assertEquals(1, metrics.get(0).samples.size());
```
k-wall, line 59: extract variable to avoid the repetition.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
k-wall: I'm not sure this is worthy of a log at info level.

mimaison: If the metric is measurable, the value is a primitive (double), so it can't be null, otherwise it can be an arbitrary type. While I don't expect null values, in this case it's easy to be safe so I agree with @k-wall, let's be defensive.

mimaison:
Thanks for the PR. 
I seems to only handle non-numeric metrics in `KafkaMetricsCollector`. You also need to do it in `YammerMetricsCollector.convert()`

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -100,4 +104,4 @@ private KafkaMetric buildNonNumericMetric(String name, String group) {
                 time);
     }
 
-}
+}
```
mimaison, line 107: Let's keep the newline

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: Or do you mean something like:
```
List<Collector.MetricFamilySamples.Sample> samples = metrics.get(0).samples;
        assertEquals(1, metrics.size());
        assertEquals("kafka_server_group_name", metrics.get(0).name);
        assertEquals(1, samples.size());
        assertEquals(1.0, samples.get(0).value, 0);
        assertTrue(samples.get(0).labelNames.contains("value"));
        }

Comment (k-wall) [reply src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java:59]: I mean extract `metrics.get(0)` as a variable.

Comment (OwenCorrigan76) [reply src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java:59]: Thanks

File: src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java (hunk 1 above)
Comment (OwenCorrigan76) [reply src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java]: Addressed

File: src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java (hunk 2 above)
Comment (OwenCorrigan76) [reply src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java]: Changed to String.valueOf(value)
LOG.info removed
`value` replaced with `name`, outputting the metric name for the label

File: src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java (hunk 4 above)
Comment (OwenCorrigan76) [reply src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java]: Value set to 0.1 like in other test assertions

File: src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java (hunk 5 above)
Comment (OwenCorrigan76) [reply src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java]: Addressed

File: src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java (hunk 6 above)
Comment (OwenCorrigan76) [reply src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java:123]: Addressed

File: src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java (hunk 2 above)
Comment (mimaison) [reply src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java]: We need to use the name of the metric attribute. Here `name` is the full metric name.
For example, if the metric is `kafka_server_app_info_commit_id`, the label should just be `commit_id` and not the full name. The attribute name can be retrieved from the MetricName object.

File: src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java [hunk 7]
```diff
@@ -110,7 +110,8 @@ String metricName(MetricName metricName) {
     }
 
     static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels) {
-        Object value = metric.metricValue();
+        Object valueObj = metric.metricValue();
+        final double value;
```
mimaison, line 114: We don't use the `final` modifier on other local variables throughout the code base. Let's keep it consistent.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -50,33 +50,25 @@ public void testCollect() {
...
-        assertEquals(1, metrics.size());
-        assertEquals("kafka_server_group_name", metrics.get(0).name);
-        assertEquals(1, metrics.get(0).samples.size());
-        assertEquals(1.0, metrics.get(0).samples.get(0).value, 1.0);
-        assertTrue(metrics.get(0).samples.get(0).labelNames.contains("value"));
-
         // Adding a metric that matches the allowlist
         collector.addMetric(buildMetric("name", "group", 1.0));
         metrics = collector.collect();
+        Collector.MetricFamilySamples getMetrics = metrics.get(0);
```
mimaison, line 56: Can we find a better variable name? Maybe `metricFamilySamples`

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -51,28 +50,27 @@ public void testCollect() {
...
+
         assertEquals(1, metrics.size());
-        assertEquals("kafka_server_group_name", metrics.get(0).name);
-        assertEquals(1, metrics.get(0).samples.size());
-        assertEquals(1.0, metrics.get(0).samples.get(0).value, 0.1);
-        assertEquals(new ArrayList<>(labels.keySet()), metrics.get(0).samples.get(0).labelNames);
-        assertEquals(new ArrayList<>(labels.values()), metrics.get(0).samples.get(0).labelValues);
+        assertEquals("kafka_server_group_name", getMetrics.name);
+        assertEquals(1, getMetrics.samples.size());
+        assertEquals(1.0, getMetrics.samples.get(0).value, 0.1);
```
k-wall, line 61: Sorry, I applied my comment yesterday to the wrong variable.  I meant to say extract `metrics.samples.get(0)` as a variable, say `serverGroupNameSamples`.   Looking again now, I actually think I'd expect a method assertMetric(..) which would validate that the expected metric exists, with the right value and labels.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -100,4 +98,28 @@ private KafkaMetric buildNonNumericMetric(String name, String group) {
...
+        List<Collector.MetricFamilySamples> metrics = collector.collect();
+        assertTrue(metrics.isEmpty());
+
+        // Adding a non-numeric metric converted
+        KafkaMetric nonNumericMetric = buildNonNumericMetric("name", "group");
+        collector.addMetric(nonNumericMetric);
+        metrics = collector.collect();
+        Collector.MetricFamilySamples getMetrics = metrics.get(0);
+
+        assertEquals(1, metrics.size());
```
k-wall, line 118: assert the size before you access the first element.

```suggestion
        assertEquals(1, metrics.size());
        Collector.MetricFamilySamples getMetrics = metrics.get(0);
```

OwenCorrigan76: Moved

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: Addresses but would like feedback on whether it could be done better

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: Changed

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: `final` removed

src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
```diff
@@ -121,14 +122,26 @@ static MetricFamilySamples convert(String name, Counter counter, Map<String, Str
...
+        Map<String, String> sanitizedLabels = labels.entrySet().stream()
+                .collect(Collectors.toMap(
+                        e -> Collector.sanitizeMetricName(e.getKey()),
+                        Map.Entry::getValue,
+                        (v1, v2) -> {
+                            throw new IllegalStateException("Unexpected duplicate key " + v1);
+                        },
+                        LinkedHashMap::new));
+
+        if (valueObj instanceof Number) {
```
k-wall, line 136:
```suggestion
        Map<String, String> sanitizedLabels = labels.entrySet().stream()
                .collect(Collectors.toMap(
                        e -> Collector.sanitizeMetricName(e.getKey()),
                        Map.Entry::getValue,
                        (v1, v2) -> {
                            throw new IllegalStateException("Unexpected duplicate key " + v1);
                        },
                        LinkedHashMap::new));

        Object valueObj = gauge.value();
        double value;
        if (valueObj instanceof Number) {
```

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -125,8 +120,17 @@ static MetricFamilySamples convert(String name, String help, KafkaMetric metric,
                             throw new IllegalStateException("Unexpected duplicate key " + v1);
```
k-wall, line 120: Sanitizing might result in collision. Hopefully it won't happen. I wonder if a WARN would be better that an exception.  I think as a user I would rather have metrics with an incomplete set of labels, rather than none. WDTY @mimaison

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -51,35 +50,87 @@ public void testCollect() {
...
+                    if (sample.value == expectedValue &&
+                            sample.labelNames.equals(new ArrayList<>(expectedLabels.keySet())) &&
+                            sample.labelValues.equals(new ArrayList<>(expectedLabels.values()))) {
+                        metricFound = true;
+                        break;
+                    }
+                }
+            }
+        }
+        assertTrue(metricFound, "Expected metric not found: " + expectedName);
```
k-wall, line 131: I don't really like how you've extracted this method.  You want the test to give good feedback when things go wrong so the developer knows precisely what he/she has broken.  The assertTrue masks the detail in an unhelpful way.


This is the kind of assert method I'd have extracted:

 https://github.com/OwenCorrigan76/metrics-reporter/pull/1

OwenCorrigan76: Implemented as suggested. Thanks for the suggestion

src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
OwenCorrigan76: Todo

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: Todo

mimaison: Yes agreed.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -22,21 +21,14 @@
...
     private final MetricConfig metricConfig = new MetricConfig();
     private final Time time = Time.SYSTEM;
-    private Map<String, String> labels;
-
-    @BeforeEach
-    public void setup() {
-        labels = new HashMap<>();
-        labels.put("key", "value");
-    }
+    private final Map<String, String> labels = Map.of("key", "value");
```
mimaison, line 28: Since you're making this change here, do you want to do the same in YammerMetricsCollectorTest? as it has the exact same code.

src/test/java/io/strimzi/kafka/metrics/YammerMetricsCollectorTest.java
```diff
@@ -106,4 +150,4 @@ public void removeMetric(String group, String name, String type) {
         KafkaYammerMetrics.defaultRegistry().removeMetric(metricName);
     }
 
-}
+}
```
mimaison, line 153: Let's keep the newline

src/test/java/io/strimzi/kafka/metrics/YammerMetricsCollectorTest.java
```diff
@@ -21,6 +24,7 @@
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
```
mimaison, line 27: Why do we need this annotation? If we assume tests need an empty metrics registry, I'd prefer explicitly clearing the registry between each test, instead of relying on the test running order.

src/test/java/io/strimzi/kafka/metrics/YammerMetricsCollectorTest.java
```diff
@@ -48,35 +53,74 @@ public void testCollect() {
...
-        metrics = collector.collect();
-        assertTrue(metrics.isEmpty());
-
         // Adding a metric that matches the allowlist
         Counter counter = newCounter("group", "name", "type");
         metrics = collector.collect();
         assertEquals(1, metrics.size());
+
+        Collector.MetricFamilySamples metricFamilySamples = metrics.get(0);
+        Collector.MetricFamilySamples.Sample serverGroupNameSamples = metricFamilySamples.samples.get(0);
```
mimaison, line 62: You want to move this after line 65 so we first check there's 1 sample before trying to access it.

src/test/java/io/strimzi/kafka/metrics/YammerMetricsCollectorTest.java
```diff
@@ -48,35 +53,74 @@ public void testCollect() {
...
 
         // Updating the value of the metric
         counter.inc(10);
         metrics = collector.collect();
         assertEquals(1, metrics.size());
-        assertEquals("kafka_server_group_name_type_count", metrics.get(0).name);
-        assertEquals(1, metrics.get(0).samples.size());
-        assertEquals(10.0, metrics.get(0).samples.get(0).value, 0.1);
+
+        Collector.MetricFamilySamples metricFamilySamples1 = metrics.get(0);
```
mimaison, line 75: No need to introduce new variables here, you can reassign the ones we created above.

mimaison (commented):
Thanks for the updates. I made another pass and left a few more comments.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: This is now refactored. Label has replaced full name and label

src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
OwenCorrigan76: Refactored

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: Refactored to use WARN instead of an exception

src/test/java/io/strimzi/kafka/metrics/YammerMetricsCollectorTest.java
OwenCorrigan76: Moved. Is it correct now?

src/test/java/io/strimzi/kafka/metrics/YammerMetricsCollectorTest.java
OwenCorrigan76: Removed

src/test/java/io/strimzi/kafka/metrics/YammerMetricsCollectorTest.java
OwenCorrigan76: Line at the end added back

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: We decided to keep as is because the ordering is clear

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -66,8 +66,7 @@ public List<MetricFamilySamples> collect() {
                 continue;
             }
             LOG.info("Kafka metric {} is allowed", name);
-            LOG.info("labels " + metricName.tags());
-            MetricFamilySamples sample = convert(name, metricName.description(), kafkaMetric, metricName.tags());
+            MetricFamilySamples sample = convert(name, metricName.description(), kafkaMetric, metricName.tags(), metricName);
```
k-wall, line 69: this method's api look weird. passing metricName and metricName.description() and metricName.tags()

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -109,24 +108,28 @@ String metricName(MetricName metricName) {
...
 
-    static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels) {
-        Object value = metric.metricValue();
-        if (!(value instanceof Number)) {
-            // Prometheus only accepts numeric metrics.
-            // Kafka gauges can have arbitrary types, so skip them for now
-            // TODO move non-numeric values to labels
-            return null;
-        }
+    static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels, MetricName metricName) {
```
k-wall, line 111:
```suggestion
    private static MetricFamilySamples convert(String name, String help, KafkaMetric metric, Map<String, String> labels, MetricName metricName) {
```

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -109,24 +108,28 @@ String metricName(MetricName metricName) {
...
         Map<String, String> sanitizedLabels = labels.entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
                         Map.Entry::getValue,
                         (v1, v2) -> {
-                            throw new IllegalStateException("Unexpected duplicate key " + v1);
+                            LOG.warn("Unexpected duplicate key" +  v1);
```
k-wall, line 117: Always use patterns when logging.  I'd suggest adding the word Ignoring as it makes it clear from the logs what the code is doing with the duplicate.

```suggestion
                            LOG.warn("Ignoring metric value duplicate key {}", v1);
```

src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
```diff
@@ -120,15 +121,28 @@ static MetricFamilySamples convert(String name, Counter counter, Map<String, Str
                 .build();
     }
 
-    static MetricFamilySamples convert(String name, Gauge<?> gauge, Map<String, String> labels) {
-        Object value = gauge.value();
-        if (!(value instanceof Number)) {
-            // Prometheus only accepts numeric metrics.
-            // Some Kafka gauges have string values (for example kafka.server:type=KafkaServer,name=ClusterId), so skip them
-            return null;
+    static MetricFamilySamples convert(String name, Gauge<?> gauge, Map<String, String> labels, MetricName metricName) {
```
k-wall, line 124:
```suggestion
    private static MetricFamilySamples convert(String name, Gauge<?> gauge, Map<String, String> labels, MetricName metricName) {
```

k-wall, line 124: Give `name` or `metricName` a more distinguishing name

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
k-wall: the reply doesn't really make sense, but I'm not too worried either way.

OwenCorrigan76: Sorry, wrong place for my last comment.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: @k-wall Does this satisfy the code smell? The only thing I'm not sure about is using `metric.metricName().description()` instead of `help`

   ```
               MetricFamilySamples sample = convert(prometheusMetricName,  kafkaMetric, metricName);

   private static MetricFamilySamples convert(String name, KafkaMetric metric, MetricName metricName) {
        Map<String, String> sanitizedLabels = metricName.tags().entrySet().stream()
                .collect(Collectors.toMap(
                        e -> Collector.sanitizeMetricName(e.getKey()),
                        Map.Entry::getValue,
                        (v1, v2) -> {
                            LOG.warn("Ignoring metric value duplicate key {}", v1);
                            return v1;
                        },
                        LinkedHashMap::new));
        Object valueObj = metric.metricValue();
        double value;
        if (valueObj instanceof Number) {
            value = ((Number) valueObj).doubleValue();
        } else {
            value = 1.0;
            String attributeName = metricName.name();
            sanitizedLabels.put(attributeName, String.valueOf(valueObj));
        }

        return new MetricFamilySamplesBuilder(Type.GAUGE, metric.metricName().description())
                .addSample(name, value, sanitizedLabels)
                .build();
    }

k-wall: > Does this satisfy the code smell? 

Yes, but rename the formal parameter `name` in the method declaration too.

> The only thing I'm not sure about is using metric.metricName().description() as help. In YammerMetricsCollector, the help argument is an empty String.

My comment was against KafkaMetricsCollector only, not YammerMetricsCollector.

src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
```diff
@@ -58,27 +59,27 @@ public List<MetricFamilySamples> collect() {
...
-                if (!config.isAllowed(name)) {
-                    LOG.info("Yammer metric {} is not allowed", name);
+                if (!config.isAllowed(prometheusMetricName)) {
+                    LOG.info("Yammer metric {} is not allowed", prometheusMetricName);
                     continue;
                 }
-                LOG.info("Yammer metric {} is allowed", name);
+                LOG.info("Yammer metric {} is allowed", prometheusMetricName);
                 Map<String, String> labels = labelsFromScope(metricName.getScope());
                 LOG.info("labels " + labels);
```
k-wall, line 70: Use placeholders in logging patterns. Two reasons why this is best practice:
- avoids pointless garbage.. in this case the runtime needs to be call map.toString() and create a new String object containing the concatenation `"labels " + labels.toString()`.   All of that is pointless garbage/cpu overhead if logging is turned down/off.
- what if the `labels`, when stringified contains the characters `{}`?  How do you know what the underlying log implementation will do?



```suggestion
                LOG.info("Yammer metric {} is allowed", prometheusMetricName);
                Map<String, String> labels = labelsFromScope(metricName.getScope());
                LOG.info("labels {} ", labels);
```

k-wall (commented):
couple of nits, but I think this is looking close now.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -59,15 +59,14 @@ public List<MetricFamilySamples> collect() {
             KafkaMetric kafkaMetric = entry.getValue();
             LOG.trace("Collecting Kafka metric {}", metricName);
 
-            String name = metricName(metricName);
+            String prometheusMetricName = metricName(metricName);
             // TODO Filtering should take labels into account
-            if (!config.isAllowed(name)) {
-                LOG.info("Kafka metric {} is not allowed", name);
+            if (!config.isAllowed(prometheusMetricName)) {
+                LOG.info("Kafka metric {} is not allowed", prometheusMetricName);
```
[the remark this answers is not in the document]
mimaison: Yeah we currently have debug log lines all over. There's a dedicated issue https://github.com/strimzi/metrics-reporter/issues/6 to clean that up.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -109,24 +108,28 @@ String metricName(MetricName metricName) {
...
-        return new MetricFamilySamplesBuilder(Type.GAUGE, help)
-                .addSample(name, ((Number) value).doubleValue(), sanitizedLabels)
+        Object valueObj = metric.metricValue();
+        double value;
+        if (valueObj instanceof Number) {
+            value = ((Number) valueObj).doubleValue();
+        } else {
+            value = 1.0;
+            String attributeName = metricName.name();
+            sanitizedLabels.put(attributeName, String.valueOf(valueObj));
```
mimaison, line 128: We probably want to sanitize the new label too.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -109,24 +108,28 @@ String metricName(MetricName metricName) {
...
+        Map<String, String> sanitizedLabels = metricName.tags().entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
                         Map.Entry::getValue,
                         (v1, v2) -> {
-                            throw new IllegalStateException("Unexpected duplicate key " + v1);
+                            LOG.warn("Ignoring metric value duplicate key {}", v1);
+                            return v1;
                         },
                         LinkedHashMap::new));
```
mimaison, line 120: It seems we have the exact same block in YammerMetricsCollector. Maybe we can extract this into a method. WDYT?

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -109,24 +108,28 @@ String metricName(MetricName metricName) {
...
-        }
-        Map<String, String> sanitizedLabels = labels.entrySet().stream()
+    private static MetricFamilySamples convert(String prometheusMetricName, KafkaMetric metric, MetricName metricName) {
+        Map<String, String> sanitizedLabels = metricName.tags().entrySet().stream()
                 .collect(Collectors.toMap(
                         e -> Collector.sanitizeMetricName(e.getKey()),
                         Map.Entry::getValue,
                         (v1, v2) -> {
-                            throw new IllegalStateException("Unexpected duplicate key " + v1);
+                            LOG.warn("Ignoring metric value duplicate key {}", v1);
```
mimaison, line 117: Should we have a test to exercise this? Running code coverage it looks like we don't test this. If you want we can tackle that in a separate PR, in that case, please open an issue to track it.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
mimaison: I'd still prefer to use the same logic in both `KafkaMetricsCollectorTest` and `YammerMetricsCollectorTest`. So either we revert this, or we do the same in `YammerMetricsCollectorTest`.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -51,35 +44,67 @@ public void testCollect() {
...
+
+        Collector.MetricFamilySamples metricFamilySamples = metrics.get(0);
+
+        assertEquals("kafka_server_group_name", metricFamilySamples.name);
+        assertEquals(1, metricFamilySamples.samples.size());
+        assertMetricFamilySample(metricFamilySamples, "kafka_server_group_name", 1.0, expectedLabels);
+    }
+
+    private void assertMetricFamilySample(Collector.MetricFamilySamples actual, String expectedSampleName, double expectedValue, Map<String, String> expectedLabels) {
+        assertEquals(expectedSampleName, actual.name, "unexpected name");
```
mimaison, line 98: We don't put messages in other assertions in this project. I think we can drop them from here to be consistent.

src/test/java/io/strimzi/kafka/metrics/YammerMetricsCollectorTest.java
```diff
@@ -48,35 +48,73 @@ public void testCollect() {
...
+        Collector.MetricFamilySamples metricFamilySamples = metrics.get(0);
+
+        assertEquals("kafka_server_group_name_type", metricFamilySamples.name);
+        assertEquals(1, metricFamilySamples.samples.size());
+        assertMetricFamilySample(metricFamilySamples, "kafka_server_group_name_type", 1.0, expectedTags);
+    }
+
+    private void assertMetricFamilySample(Collector.MetricFamilySamples actual, String expectedSampleName, double expectedValue, Map<String, String> expectedTags) {
+        assertEquals(expectedSampleName, actual.name, "unexpected name");
+        assertEquals(1, actual.samples.size(), "unexpected number of samples");
```
mimaison, line 109: It looks like we already checked have these 2 assertions on line 102-103. Why are we checking these twice?

mimaison: Also this is only called from a single place, so maybe this can be inlined?

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -51,35 +44,67 @@ public void testCollect() {
...
+        metrics = collector.collect();
+
+        Map<String, String> expectedLabels = new LinkedHashMap<>(labels);
+        expectedLabels.put("name", nonNumericValue);
+        assertEquals(1, metrics.size());
+
+        Collector.MetricFamilySamples metricFamilySamples = metrics.get(0);
+
+        assertEquals("kafka_server_group_name", metricFamilySamples.name);
+        assertEquals(1, metricFamilySamples.samples.size());
```
mimaison, line 93: These 2 assertions are checked again in `assertMetricFamilySample()`. Why are we checking them twice?

mimaison (commented):
Thanks for the updates. I made a another pass and left a few comments.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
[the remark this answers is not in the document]
OwenCorrigan76: Will look at this in https://github.com/strimzi/metrics-reporter/issues/6

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: @mimaison Does this suffice in sanitizing the new label?
` sanitizedLabels.put(Collector.sanitizeMetricName(attributeName), String.valueOf(valueObj));`

mimaison: Yes let's use the same method we use on the other labels

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: @mimaison Are we testing whether the warning is logged or are we testing if there are duplicate keys? A bit unsure. Thanks

mimaison: I'm not sure we necessarily need to check the log line is emitted, but we can check the behavior, i.e. what happens if we add a metric that would cause duplicated labels?

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: Using a helper method of sorts? Where would be the best place to have the extracted method that would be called from both Classes?

mimaison: While not perfect, `MetricFamilySamplesBuilder` could do.

OwenCorrigan76: Sorry, can you elaborate on `MetricFamilySamplesBuilder could do`.

mimaison: I meant we could put that new method in `MetricFamilySamplesBuilder`

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: `Collector.sanitizeMetricName` used

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: New `sanitizeLabels` method added to `MetricFamilySamplesBuilder` and called in `convert` methods

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: New issue created for this: https://github.com/strimzi/metrics-reporter/issues/22

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: All messages removed

src/test/java/io/strimzi/kafka/metrics/YammerMetricsCollectorTest.java
OwenCorrigan76: Duplicates removed and changed to inline

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: Duplicates removed

src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
OwenCorrigan76: Thanks for the explanation. Done

k-wall (approved):
lgtm, thank you @OwenCorrigan76

scholzj (commented):
It would be great if you could provide a bit more details in the Pr description about what and how it does.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -59,15 +57,14 @@ public List<MetricFamilySamples> collect() {
...
-            if (!config.isAllowed(name)) {
-                LOG.info("Kafka metric {} is not allowed", name);
+            if (!config.isAllowed(prometheusMetricName)) {
+                LOG.info("Kafka metric {} is not allowed", prometheusMetricName);
                 continue;
             }
-            LOG.info("Kafka metric {} is allowed", name);
-            LOG.info("labels " + metricName.tags());
-            MetricFamilySamples sample = convert(name, metricName.description(), kafkaMetric, metricName.tags());
+            LOG.info("Kafka metric {} is allowed", prometheusMetricName);
```
scholzj, line 66: I know this was already here. But do we really need this on INFO level? Maybe DEBUG would be enough?

src/main/java/io/strimzi/kafka/metrics/MetricFamilySamplesBuilder.java
```diff
@@ -6,18 +6,23 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Helper class to convert Kafka metrics into the Prometheus format.
  */
 public class MetricFamilySamplesBuilder {
 
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaMetricsCollector.class.getName());
```
mimaison, line 25: `KafkaMetricsCollector` -> `MetricFamilySamplesBuilder`

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
mimaison: https://github.com/strimzi/metrics-reporter/pull/21/files#r1644743277 has not been addressed

src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java
```diff
@@ -58,27 +58,27 @@ public List<MetricFamilySamples> collect() {
...
-                    LOG.info("Yammer metric {} is not allowed", name);
+                if (!config.isAllowed(prometheusMetricName)) {
+                    LOG.info("Yammer metric {} is not allowed", prometheusMetricName);
                     continue;
                 }
-                LOG.info("Yammer metric {} is allowed", name);
+                LOG.info("Yammer metric {} is allowed", prometheusMetricName);
                 Map<String, String> labels = labelsFromScope(metricName.getScope());
-                LOG.info("labels " + labels);
+                LOG.info("labels {} ", labels);
```
scholzj, line 69: Should this have some formatting? Or should it be removed? It looks like some leftover debug metric.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
mimaison: I think it would be more idiomatic to either check the size in `assertMetricFamilySample()`, or pass the unique sample (`metricFamilySamples.samples.get(0)`) to `assertMetricFamilySample()`.

It's weird to check the sample size is one and still pass the list of samples to `assertMetricFamilySample()`.

Same in `YammerMetricsCollectorTest`.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -90,8 +113,8 @@ private KafkaMetric buildMetric(String name, String group, double value) {
                 time);
     }
 
-    private KafkaMetric buildNonNumericMetric(String name, String group) {
-        Gauge<String> measurable = (config, now) -> "hello";
+    private KafkaMetric buildNonNumericMetric(String name, String group, String nonNumericValue) {
```
mimaison, line 116: Nit: this can be just `value`. It's type `String`, so we know it's not numeric.

mimaison (commented):
Thanks for the updates, I think we're almost there. I left a few small suggestions.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
mimaison: Yes we plan to address this in https://github.com/strimzi/metrics-reporter/issues/6

scholzj: It is changed in this PR and people have to review this, so it can be easily addressed right now I guess?

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: Sorry, I missed this comment somehow. To keep uniformity, I will revert `KafkaMetricsReporter` to the former logic. 
In trying to implement `private final LinkedHashMap<String, String> tags = new LinkedHashMap<>(Map.of("key", "value", "key2", "value2"));
`, while the LinkedHashMap provided ordering guarantee, I think the Map.of method creates an immutable map, and using it directly to initialize a mutable LinkedHashMap might cause issues in some cases, causing the test to fail sometimes.

src/main/java/io/strimzi/kafka/metrics/MetricFamilySamplesBuilder.java
OwenCorrigan76: Silly mistake. Fixed

OwenCorrigan76:
@scholzj Is the PR description better? Thanks

scholzj (approved):

k-wall (approved):

mimaison (approved):
Thanks for implementing this feature and addressing all the comments!

There are still a couple of things not quite right in the tests (for example we access the first item before checking the collection size is 1 in [`assertMetricFamilySample()`](https://github.com/strimzi/metrics-reporter/pull/21/files#diff-e7c561ab40b7a00f9bebb866d376a6e907c485680bc59ad93aac384738e6107eR108-R110)) but we can address them in follow up PRs.

LGTM

Issue #24, opened 2024-06-22
Error reading assemblies: No assembly descriptors found


antonio-pedro99:
I am trying to set up the project locally, and facing the following error

```bash
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.2-beta-5:single (default-cli) on project metrics-reporter: Error reading assemblies: No assembly descriptors found. -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

From where can I get the assembly descriptor?

mimaison:
Hi @antonio-pedro99,

The initial import used the maven assembly plugin to generate a JAR with dependencies. It looks like the plugin was removed in https://github.com/strimzi/metrics-reporter/commit/9b8756d7643224614b9f78d2b49e54de6faabd22 by @scholzj. He did keep the version definition but commented it out: https://github.com/strimzi/metrics-reporter/blob/main/pom.xml#L110.

I'm not sure if it was done for a specific reason or if it's only temporary. In the meantime, you can uncomment the version definition and add the following to the `plugins` section of the `pom.xml` file:
```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-assembly-plugin</artifactId>
    <version>${maven.assembly.version}</version>
    <executions>
        <execution>
            <phase>package</phase>
            <goals><goal>single</goal></goals>
        </execution>
    </executions>
    <configuration>
        <descriptorRefs>
            <descriptorRef>jar-with-dependencies</descriptorRef>
        </descriptorRefs>
    </configuration>
</plugin>
```

scholzj:
It is working on the CIs. So I'm not sure this is a problem of something missing but more some kind of environment issue or some issue with what is being run?

mimaison:
Is the CI running `assembly:single`?

scholzj:
I guess it does not. And I'm not sure what would such command be even expected to do TBH.

scholzj:
I know what the assembly plugin does, but I never saw it used in that way. I think it should be removed from the README.

mimaison:
[...]

PR #25, opened 2024-06-25
Rewrite to use the new Prometheus Java client

mimaison:
This introduces some changes:

- Quantiles names are now serialized without trailing zeros. So we now have labels like `quantile="0.5"`. Previously the label was `quantile="0.50"`.
- I've added `sum` series to Summary metrics. They now have quantiles, count and sum. Sum series have the `_sum` suffix.
- Removed `_count` suffix from Counter metrics. As per the [Counter documentation](https://prometheus.github.io/client_java/getting-started/metric-types/#counter), Counter metrics should use the `_total` suffix. Previously we were injecting `_count` and Prometheus was adding `_total`. So now we have `kafka_server_kafka_server_brokertopicmetrics_bytesinpersec_total` while before we had `kafka_server_kafka_server_brokertopicmetrics_bytesinpersec_count_total`.
- The new client has a dedicated metric type of non-numeric metrics, `InfoSnapshot`. Under the covers this still emits a Gauge metric with a value set to 1, like we do today. This is because Prometheus does not support non-numeric metrics. It's preferable to use the dedicated type, however this new type uses the `_info` suffix. So now we have `kafka_server_kafka_server_kafkaserver_clusterid_info{ClusterId="mHGu-1meQOSC_RpQHdCpOw"}` while before we had `kafka_server_kafka_server_kafkaserver_clusterid{ClusterId="mHGu-1meQOSC_RpQHdCpOw"}`. The suffix helps identify metrics whose values should not be graphed (as they are constant) and where the useful information is in the labels instead.
- Gauge metrics don't have a suffix anymore, as per the [OpenMetrics specification](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#suffixes). The previous Prometheus client was injecting a `_total` suffix. We now have `kafka_server_txn_marker_channel_metrics_response`, while previously we had `kafka_server_txn_marker_channel_metrics_request_total`.
- The new client does not insert a trailing comma after the last label. We now have `kafka_server_alter_partition_metrics_connection_close_rate{BrokerId="0"}`, while we previously had `kafka_server_alter_partition_metrics_connection_close_rate{BrokerId="0",}`.

The new client allows exposing metrics in the OpenMetrics format and serialized with Protobuf. When Prometheus scrapes the endpoint, it negotiates the format using the `Accept` HTTP header. You can check different formats using the `debug` query parameter on the `GET /metrics` endpoint. For example:

- `/metrics?debug=openmetrics`: View OpenMetrics text format.
- `/metrics?debug=text`: View Prometheus text format.
- `/metrics?debug=prometheus-protobuf`: View a text representation of the Prometheus protobuf format.

Related resources:
- http://prometheus.github.io/client_java/
- https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md


src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -90,36 +73,60 @@ public void removeMetric(KafkaMetric metric) {
...
-        String name = metricName.name()
-                .replace('.', '_')
-                .replace('-', '_')
+    /**
+     * Called when the Prometheus server scrapes metrics.
+     * @return metrics that match the configured allow list
+     */
+    @Override
+    public MetricSnapshots collect() {
+        Map<String, GaugeSnapshot.Builder> gaugeBuilders = new HashMap<>();
```
mimaison, line 82: Unfortunately the `MetricSnapshotBuilder` interface does not have a `build()` method. I opened a [PR](https://github.com/prometheus/client_java/pull/969) against [prometheus/client_java](https://github.com/prometheus/client_java) to change that.

In the meantime, I think we have to keep a collection per metric type.

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+                .build();
+    }
+
+    /**
+     * Convert a Number into a datapoint for a {@link CounterSnapshot} metric
+     * @param number The value
+     * @param labels Labels associated with the datapoint
+     * @return The {@link CounterSnapshot.CounterDataPointSnapshot} datapoint
+     */
+    public static CounterSnapshot.CounterDataPointSnapshot convert(Number number, Labels labels) {
```
mimaison, line 59: I'm not super happy with the various methods especially as `convert(double value, Labels labels)` and `convert(Number number, Labels labels)` shadow each other. Ideally I wanted to keep this file "Yammer free" but I ended up having to use Timer and Histogram.

So maybe we should tweak some of these `convert()` methods to accept the Yammer and Kafka metric types.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -90,36 +73,60 @@ public void removeMetric(KafkaMetric metric) {
...
+            snapshots.add(builder.build());
+        }
+        for (InfoSnapshot.Builder builder : infoBuilders.values()) {
+            snapshots.add(builder.build());
+        }
+        return new MetricSnapshots(snapshots);
+    }
+
+    private String metricName(MetricName metricName) {
+        return PrometheusNaming
```
mimaison, line 120: I'm not sure I fully understand how to use the various methods in `PrometheusNaming` especially `sanitizeMetricName()` and `prometheusName()`.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -69,17 +70,20 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
...
+    private final PrometheusRegistry registry;
 
     /**
      * Constructor.
      *
      * @param props the configuration properties.
+     * @param registry the metrics registry
      */
-    public PrometheusMetricsReporterConfig(Map<?, ?> props) {
+    public PrometheusMetricsReporterConfig(Map<?, ?> props, PrometheusRegistry registry) {
```
mimaison, line 81: `PrometheusRegistry` has no methods to clear it. For this reason, I introduced constructors to inject a registry instance. This simplifies testing.

src/main/java/io/strimzi/kafka/metrics/KafkaPrometheusMetricsReporter.java
```diff
@@ -30,42 +30,57 @@ public class KafkaPrometheusMetricsReporter implements MetricsReporter {
...
 
+    /**
+     * Constructor
+     */
+    public KafkaPrometheusMetricsReporter() {
+        registry = PrometheusRegistry.defaultRegistry;
+    }
+
+    // for testing
+    KafkaPrometheusMetricsReporter(PrometheusRegistry registry) {
```
mimaison, line 48: `PrometheusRegistry` has no methods to clear it. For this reason, I introduced constructors to inject a registry instance. This simplifies testing. 
Same in `YammerPrometheusMetricsReporter`

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
mimaison, line 81: It's not ideal to have a `registry` field here. But it's required when start the HTTP server.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -14,107 +19,109 @@
...
     private final Time time = Time.SYSTEM;
-    private Map<String, String> labels;
+    private Map<String, String> tagsMap;
+    private Labels labels;
 
     @BeforeEach
     public void setup() {
-        labels = new HashMap<>();
-        labels.put("key", "value");
+        tagsMap = new LinkedHashMap<>();
```
mimaison, line 38: The setup of the 2 Collector tests is still a bit wonky with the `labels`/`tagsMap` fields. I'll see if I can tidy this up a little bit.

pom.xml
```diff
@@ -309,6 +309,23 @@
           </archive>
         </configuration>
       </plugin>
+      <!-- Temporarily re-add the assembly plugin before we fix the packaging -->
```
mimaison, line 312: This will be addressed in https://github.com/strimzi/metrics-reporter/issues/24

pom.xml
```diff
@@ -169,17 +169,17 @@
 
     <dependency>
       <groupId>io.prometheus</groupId>
-      <artifactId>simpleclient</artifactId>
+      <artifactId>prometheus-metrics-model</artifactId>
```
k-wall, line 172: Should we just import the BOM https://central.sonatype.com/artifact/io.prometheus/prometheus-metrics-bom

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+                .labels(labels)
+                .quantiles(quantiles(histogram))
+                .count(histogram.count())
+                .sum(histogram.sum())
+                .build();
+    }
+
+    private static Quantiles quantiles(Sampling samplingMetric) {
+        Quantiles.Builder quantilesBuilder = Quantiles.builder();
+        for (double quantile : Arrays.asList(0.50, 0.75, 0.95, 0.98, 0.99, 0.999)) {
```
k-wall, line 98: extract constant

pom.xml
scholzj: We try to avoid that. It makes it hard to override dependencies etc.

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Convert a non numeric metric into a datapoint for a {@link InfoSnapshot} metric
+     * @param value The value of the datapoint
+     * @param labels Labels associated with the datapoint
+     * @param metricName The name of the new label
+     * @return The {@link InfoSnapshot.InfoDataPointSnapshot} datapoint
+     */
+    public static InfoSnapshot.InfoDataPointSnapshot convert(Object value, Labels labels, String metricName) {
+        labels = labels.add(PrometheusNaming.sanitizeLabelName(metricName), String.valueOf(value));
```
k-wall, line 34: do we want to mutate the caller's labels?

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
k-wall: if you had  sum and count parameters that were Suppliers and a value Function, you could hide yammerness from this class.

mimaison: I shuffled a few things around so this class does not depend on Yammer anymore.

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
mimaison: `Labels.add()` returns a new instance, but yeah we don't want to reassign `labels`. Fixed.

k-wall (approved):
lgtm

.checkstyle/suppressions.xml
```diff
@@ -5,5 +5,8 @@
         "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 
 <suppressions>
-  <!-- No suppressions here -->
+  <suppress checks="CyclomaticComplexity" files="YammerMetricsCollector.java"/>
+  <suppress checks="NPathComplexity" files="YammerMetricsCollector.java"/>
+  <suppress checks="JavaNCSS" files="YammerMetricsCollector.java"/>
+  <suppress checks="ClassFanOutComplexity" files="YammerMetricsCollector.java"/>
```
scholzj, line 11: Should you use annotations to suppress these instead of suppressing it here? That is what we normally do -> it is better coupled with the code and easier to see and maintain.

mimaison: Right, if it's what Strimzi usually does, yes let's use annotations. Done

scholzj (approved):

mimaison:
I rebased on `main` and removed my `pom.xml` changes related to packaging as it's been resolved in #27

PR #28, opened 2024-06-30
Fix pushing of artifacts to Nexus

scholzj:
### Type of change

- Bugfix

### Description

Some changes at Sonatype broker the way we push the JAR artifacts to Sonatype Nexus. This PR fixes it by updating the Sonatype plugin. It also updates the script we use for it to make it fail in case of problems.


mimaison (approved):
LGTM

see-quick (approved):

Issue #22, opened 2024-06-19
Test ignoring metric value duplicate key


OwenCorrigan76:
We should have a test to exercise  what happens if we add a metric that would cause duplicated labels? This is in the `KafkaMetricsCollector` in the `convert` method and in the YammerMetricsCollector` in the `convert` method.
`LOG.warn("Ignoring metric value duplicate key {}", v1); `

OwenCorrigan76:
@mimaison @k-wall 
The logic that handles non-numeric metrics is in the [DataPointSnapshotBuilder](https://github.com/strimzi/metrics-reporter/blob/main/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java) class. [Here](https://github.com/strimzi/metrics-reporter/blob/main/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java#L28) is where a new label with a metric value is created.
Currently we unconditionally add our new label without checking if a label with the same name exists already.
Is this the correct desired behaviour or should we also consider either of the following posibilities:

* Do nothing if a label with this name already exists
* Add a WARN log if it happens 

In the previous [iteration](https://github.com/OwenCorrigan76/metrics-reporter/blob/non-numeric_values/src/main/java/io/strimzi/kafka/metrics/MetricFamilySamplesBuilder.java#L76-L85) for version 0.16.0 we had the following warning:
`LOG.warn("Ignoring metric value duplicate key {}", v1)`
Should we have something like this again? 

Two question I have.
* What are the circumstances that would lead to multiple labels with the same name existing? Would this be a user who would do this?
* What are the ramifications if we do add another label with the same name as an existing label which could potentially happen with the current logic?

OwenCorrigan76:
II've written this to add the warning back in. WDYT?
 ```public static InfoSnapshot.InfoDataPointSnapshot infoDataPoint(Labels labels, Object value, String metricName) {
        Map<String, String> labelMap = new HashMap<>();
        labels.forEach(label -> labelMap.put(label.getName(), label.getValue()));

        String sanitizedMetricName = PrometheusNaming.sanitizeLabelName(metricName);
        String valueString = String.valueOf(value);
        
        if (labelMap.containsKey(sanitizedMetricName)) {
            LOG.warning("Ignoring metric value duplicate key: " + sanitizedMetricName);
            labelMap.put(sanitizedMetricName, valueString);
        } else {
[...]

PR #29, opened 2024-07-08
Add warning log when duplicate leables are created or detected

OwenCorrigan76:
The logic that handles non-numeric metrics is in the [DataPointSnapshotBuilder](https://github.com/strimzi/metrics-reporter/blob/main/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java) class. [Here](https://github.com/strimzi/metrics-reporter/blob/main/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java#L28) is where a new label with a metric value is created.
Currently we unconditionally add our new label without checking if a label with the same name exists already.
This PR logs a warning if duplicate labes are detected rather than throwing an `IllegalArgumentException`

### DIFF
--- a/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
+++ b/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
@@ -12,10 +12,13 @@ import io.prometheus.metrics.model.snapshots.PrometheusNaming;
 import io.prometheus.metrics.model.snapshots.Quantiles;
 import io.prometheus.metrics.model.snapshots.SummarySnapshot;
 
+import java.util.logging.Logger;
 /**
  * Builder methods to convert Kafka metrics into Prometheus metrics
  */
 public class DataPointSnapshotBuilder {
+
+    private static final Logger LOG = Logger.getLogger(DataPointSnapshotBuilder.class.getName());
 
     /**
      * Create a datapoint for a {@link InfoSnapshot} metric
@@ -24,8 +27,15 @@ public class DataPointSnapshotBuilder {
      * @param metricName The name of the new label
      * @return The {@link InfoSnapshot.InfoDataPointSnapshot} datapoint
      */
+
     public static InfoSnapshot.InfoDataPointSnapshot infoDataPoint(Labels labels, Object value, String metricName) {
-        Labels newLabels = labels.add(PrometheusNaming.sanitizeLabelName(metricName), String.valueOf(value));
+        String sanitizedMetricName = PrometheusNaming.sanitizeLabelName(metricName);
+        Labels newLabels = labels;
+        if (labels.contains(sanitizedMetricName)) {
+            LOG.warning("Ignoring metric value duplicate key: " + sanitizedMetricName);
+        } else {
+            newLabels = labels.add(sanitizedMetricName, String.valueOf(value));
+        }
         return InfoSnapshot.InfoDataPointSnapshot.builder()
                 .labels(newLabels)
                 .build();
--- a/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
+++ b/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
@@ -11,14 +11,15 @@ import io.prometheus.metrics.model.snapshots.Labels;
 import io.prometheus.metrics.model.snapshots.PrometheusNaming;
 import io.prometheus.metrics.model.snapshots.Quantiles;
 import io.prometheus.metrics.model.snapshots.SummarySnapshot;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import java.util.logging.Logger;
 /**
  * Builder methods to convert Kafka metrics into Prometheus metrics
  */
 public class DataPointSnapshotBuilder {
 
-    private static final Logger LOG = Logger.getLogger(DataPointSnapshotBuilder.class.getName());
+    private static final Logger LOG = LoggerFactory.getLogger(DataPointSnapshotBuilder.class.getName());
 
     /**
      * Create a datapoint for a {@link InfoSnapshot} metric
@@ -29,12 +30,12 @@ public class DataPointSnapshotBuilder {
      */
 
     public static InfoSnapshot.InfoDataPointSnapshot infoDataPoint(Labels labels, Object value, String metricName) {
-        String sanitizedMetricName = PrometheusNaming.sanitizeLabelName(metricName);
+        String newLabelName = PrometheusNaming.sanitizeLabelName(metricName);
         Labels newLabels = labels;
-        if (labels.contains(sanitizedMetricName)) {
-            LOG.warning("Ignoring metric value duplicate key: " + sanitizedMetricName);
+        if (labels.contains(newLabelName)) {
+            LOG.warn("Ignoring metric value duplicate key: {} from metric: {}",  newLabelName, metricName);
         } else {
-            newLabels = labels.add(sanitizedMetricName, String.valueOf(value));
+            newLabels = labels.add(newLabelName, String.valueOf(value));
         }
         return InfoSnapshot.InfoDataPointSnapshot.builder()
                 .labels(newLabels)
--- a/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
+++ b/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
@@ -28,7 +28,6 @@ public class DataPointSnapshotBuilder {
      * @param metricName The name of the new label
      * @return The {@link InfoSnapshot.InfoDataPointSnapshot} datapoint
      */
-
     public static InfoSnapshot.InfoDataPointSnapshot infoDataPoint(Labels labels, Object value, String metricName) {
         String newLabelName = PrometheusNaming.sanitizeLabelName(metricName);
         Labels newLabels = labels;
--- a/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
+++ b/src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
@@ -32,8 +32,12 @@ public class DataPointSnapshotBuilder {
     public static InfoSnapshot.InfoDataPointSnapshot infoDataPoint(Labels labels, Object value, String metricName) {
         String newLabelName = PrometheusNaming.sanitizeLabelName(metricName);
         Labels newLabels = labels;
-        if (labels.contains(newLabelName)) {
-            LOG.warn("Ignoring metric value duplicate key: {} from metric: {}",  newLabelName, metricName);
+        String existingValue = labels.get(newLabelName);
+        if (existingValue != null) {
+            if (!String.valueOf(value).equals(existingValue)) {
+                LOG.warn("Unable to add new label because of duplicate key: {} with value: {} from metric: {}",
+                        newLabelName, value, metricName);
+            }
         } else {
             newLabels = labels.add(newLabelName, String.valueOf(value));
         }
--- a/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
@@ -96,7 +96,7 @@ public class KafkaMetricsCollector implements MultiCollector {
                 continue;
             }
             LOG.info("Kafka metric {} is allowed", prometheusMetricName);
-            Labels labels = labelsFromTags(metricName.tags());
+            Labels labels = labelsFromTags(metricName.tags(), metricName.name());
 
             Object valueObj = kafkaMetric.metricValue();
             if (valueObj instanceof Number) {
@@ -124,7 +124,7 @@ public class KafkaMetricsCollector implements MultiCollector {
                 .toLowerCase(Locale.ROOT);
     }
 
-    protected static Labels labelsFromTags(Map<String, String> tags) {
+    static Labels labelsFromTags(Map<String, String> tags, String metricName) {
         Labels.Builder builder = Labels.builder();
         Set<String> labelNames = new HashSet<>();
         for (Map.Entry<String, String> label : tags.entrySet()) {
@@ -132,7 +132,7 @@ public class KafkaMetricsCollector implements MultiCollector {
             if (labelNames.add(newLabelName)) {
                 builder.label(newLabelName, label.getValue());
             } else {
-                LOG.warn("Ignoring duplicate label key: {} with value: {}", newLabelName, label.getValue());
+                LOG.warn("Ignoring duplicate label key: {} with value: {} from metric: {} ", newLabelName, label.getValue(), metricName);
             }
         }
         return builder.build();
--- a/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
+++ b/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
@@ -19,9 +19,11 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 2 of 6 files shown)


OwenCorrigan76:
@mimaison I've refactored the infoDataPoint to use the `contains` method. This takes care of the newly created labels right?
Is it in `gaugeDataPoint` that I should  handle already existing lables?

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
```diff
@@ -12,20 +12,30 @@
 import io.prometheus.metrics.model.snapshots.Quantiles;
 import io.prometheus.metrics.model.snapshots.SummarySnapshot;
 
+import java.util.logging.Logger;
```
mimaison, line 15: We should use `org.slf4j.Logger` instead of `java.util.logging.Logger`

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
```diff
@@ -12,20 +12,30 @@
...
      * Create a datapoint for a {@link InfoSnapshot} metric
      * @param labels The labels associated with the datapoint
      * @param value The value to insert as a label
      * @param metricName The name of the new label
      * @return The {@link InfoSnapshot.InfoDataPointSnapshot} datapoint
      */
+
     public static InfoSnapshot.InfoDataPointSnapshot infoDataPoint(Labels labels, Object value, String metricName) {
-        Labels newLabels = labels.add(PrometheusNaming.sanitizeLabelName(metricName), String.valueOf(value));
+        String sanitizedMetricName = PrometheusNaming.sanitizeLabelName(metricName);
```
mimaison, line 32: I think naming this `newLabelName` would better describe its use.

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
```diff
@@ -12,20 +12,30 @@
...
+        Labels newLabels = labels;
+        if (labels.contains(sanitizedMetricName)) {
+            LOG.warning("Ignoring metric value duplicate key: " + sanitizedMetricName);
```
mimaison, line 35: It's probably worth print a bit more information such as the metric name so operators can identify the culprit.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -108,6 +109,19 @@ public void testCollectNonNumericMetric() {
         assertEquals(expectedLabels, snapshot.getDataPoints().get(0).getLabels());
     }
 
+    @Test
+    public void testLabelNameSanitizing() {
```
mimaison, line 113: We want to avoid tests that only exercise code from dependencies. This test only validates `PrometheusNaming.sanitizeLabelName()` and it does not check anything from our metric reporter.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -108,6 +109,19 @@ public void testCollectNonNumericMetric() {
...
+        String labelName = PrometheusNaming.sanitizeLabelName("k-1");
+        assertEquals("k_1", labelName);
+    }
+
+    @Test
+    public void testLabelDuplicateWarning() {
+        Labels labels = Labels.builder().label("k1", "v1").build();
+        InfoSnapshot.InfoDataPointSnapshot snapshot = DataPointSnapshotBuilder.infoDataPoint(labels, "v1", "k1");
```
mimaison, line 121: We should use a different value than `v1` here. Otherwise this does not assert we did not replace the value.

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
mimaison: Also we should not use concatenation in log lines but instead use interpolation with the `{}` syntax which is possible with the slf4j `Logger`

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
```diff
@@ -108,6 +109,19 @@ public void testCollectNonNumericMetric() {
...
     }
 
+    @Test
+    public void testLabelNameSanitizing() {
+        String labelName = PrometheusNaming.sanitizeLabelName("k-1");
+        assertEquals("k_1", labelName);
+    }
+
+    @Test
+    public void testLabelDuplicateWarning() {
```
mimaison, line 119: I think we should put this test in a new class `DataPointSnapshotBuilderTest`

mimaison (commented):
Thanks for the PR. I left a few suggestions. This also does not currently address the first issue identified in https://github.com/strimzi/metrics-reporter/issues/22#issuecomment-2210724766

mimaison:
> @mimaison I've refactored the infoDataPoint to use the `contains` method. This takes care of the newly created labels right? Is it in `gaugeDataPoint` that I should handle already existing lables?

No, it's in the 2 places where we create `Labels` objects:
- https://github.com/strimzi/metrics-reporter/blob/main/src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java#L125-L131
- https://github.com/strimzi/metrics-reporter/blob/main/src/main/java/io/strimzi/kafka/metrics/YammerMetricsCollector.java#L146-L157

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
OwenCorrigan76: That's why I couldn't use interpolation with `{}`. I was using the java.util logger. Changed to `sl4j`

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
OwenCorrigan76: Changed to 'newLabelName`

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
OwenCorrigan76: I appended `from metric; metricName`. Is that sufficient. 
`{}` interpolation now used

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: Removed

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: Yes, I missed that. Changed to just `value`.

src/test/java/io/strimzi/kafka/metrics/KafkaMetricsCollectorTest.java
OwenCorrigan76: New `DataPointSnapshotBuilderTest` class created and this test is moved to there

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
```diff
@@ -11,21 +11,32 @@
...
      * @param metricName The name of the new label
      * @return The {@link InfoSnapshot.InfoDataPointSnapshot} datapoint
      */
+
     public static InfoSnapshot.InfoDataPointSnapshot infoDataPoint(Labels labels, Object value, String metricName) {
-        Labels newLabels = labels.add(PrometheusNaming.sanitizeLabelName(metricName), String.valueOf(value));
+        String newLabelName = PrometheusNaming.sanitizeLabelName(metricName);
+        Labels newLabels = labels;
+        if (labels.contains(newLabelName)) {
+            LOG.warn("Ignoring metric value duplicate key: {} from metric: {}",  newLabelName, metricName);
```
mimaison, line 36: I wonder if it would be worth checking the value of the existing label and only print the warning if the value is different from `value`. WDYT?

Can we also improve the message? We're not ignoring the metric value, we're unable to add a new label.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -122,10 +124,16 @@ private String metricName(MetricName metricName) {
                 .toLowerCase(Locale.ROOT);
     }
 
-    private static Labels labelsFromTags(Map<String, String> tags) {
+    protected static Labels labelsFromTags(Map<String, String> tags) {
```
mimaison, line 127: We don't need `protected`, we don't expect subclasses

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
```diff
@@ -122,10 +124,16 @@ private String metricName(MetricName metricName) {
...
         Labels.Builder builder = Labels.builder();
+        Set<String> labelNames = new HashSet<>();
         for (Map.Entry<String, String> label : tags.entrySet()) {
-            builder.label(PrometheusNaming.sanitizeLabelName(label.getKey()), label.getValue());
+            String newLabelName = PrometheusNaming.sanitizeLabelName(label.getKey());
+            if (labelNames.add(newLabelName)) {
+                builder.label(newLabelName, label.getValue());
+            } else {
+                LOG.warn("Ignoring duplicate label key: {} with value: {}", newLabelName, label.getValue());
```
mimaison, line 135: If wonder if it's worth passing the metric name so the message helps identify the impacted metric.

Same in `YammerMetricsCollector`.

src/test/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilderTest.java
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DataPointSnapshotBuilderTest {
+
+    @Test
+    public void testNewLabelDuplicateWarning() {
+        Labels labels = Labels.builder().label("k1", "v1").label("k2", "v2").build();
+        InfoSnapshot.InfoDataPointSnapshot snapshot = DataPointSnapshotBuilder.infoDataPoint(labels, "value", "k1");
```
mimaison, line 18: I think it would be interesting to also test with labels that are not equal but collide when turned into Prometheus labels, for example `k-1` and `k_1`

src/test/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilderTest.java
```diff
@@ -0,0 +1,22 @@
...
+import io.prometheus.metrics.model.snapshots.InfoSnapshot;
+import io.prometheus.metrics.model.snapshots.Labels;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DataPointSnapshotBuilderTest {
+
+    @Test
+    public void testNewLabelDuplicateWarning() {
```
mimaison, line 16: I think a better name would be something like `testCollidingNewLabelIsIgnored`

mimaison (commented):
Thanks for the updates! It looks good overall. I left a few minor suggestions.

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: @mimaison I'm calling labelsFromTags in the test:
 @Test
    public void testLabelsFromTags() {
        Map<String, String> tags = new LinkedHashMap<>();
        tags.put("k-1", "v1");
        tags.put("k_1", "v2");

        Labels labels = KafkaMetricsCollector.labelsFromTags(tags);
        assertEquals("k_1", PrometheusNaming.sanitizeLabelName("k-1"));
        assertEquals("v1", labels.get("k_1"));
        assertEquals(1, labels.size());
    }
Is there another way around this? Thanks

OwenCorrigan76: Is there another way around this? Thanks

mimaison: > I'm calling labelsFromTags in the test
How is this an issue?

Have you tried removing `protected` and see what happens? 
You may want to review https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

OwenCorrigan76: Yes it goes to package-private. I misunderstood your comment. I thought we wanted to keep it as `private`.

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
OwenCorrigan76: Refactored. Is the new message ok?

src/main/java/io/strimzi/kafka/metrics/KafkaMetricsCollector.java
OwenCorrigan76: Metric name passed and added to the LOG

src/test/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilderTest.java
OwenCorrigan76: Just like in the other tests?
`assertEquals("k_1", PrometheusNaming.sanitizeLabelName("k-1"));`

src/test/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilderTest.java
OwenCorrigan76: Changed to `testCollidingNewLabelIsIgnored`

src/test/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilderTest.java
mimaison: No, I mean using label names in `Labels.builder().label()` and `DataPointSnapshotBuilder.infoDataPoint()` that are not equal but collide. Currently we use `k1` in both.

OwenCorrigan76: Changed to also test with labels that are not equal but collide when turned into Prometheus labels

mimaison:
@OwenCorrigan76 There's a checkstyle issue that's failing the build: https://dev.azure.com/cncf/strimzi/_build/results?buildId=174948&view=logs&j=441a6687-78c7-548e-9324-bade41cb68f6&t=cce61017-8585-548c-ef2e-c6a3bc420f66. Can you fix it? Thanks

OwenCorrigan76:
Sorry, didn't do a `mvn compile` .Fixing now.

mimaison (approved):
Thanks for the PR! LGTM

scholzj (approved):
One nit. LGTM otherwise.

src/main/java/io/strimzi/kafka/metrics/DataPointSnapshotBuilder.java
```diff
@@ -11,21 +11,36 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(DataPointSnapshotBuilder.class.getName());
+
     /**
      * Create a datapoint for a {@link InfoSnapshot} metric
      * @param labels The labels associated with the datapoint
      * @param value The value to insert as a label
      * @param metricName The name of the new label
      * @return The {@link InfoSnapshot.InfoDataPointSnapshot} datapoint
      */
+
```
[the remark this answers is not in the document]
OwenCorrigan76: @scholzj Removed extra unwanted line

PR #20, opened 2024-04-26
Address SpotBugs issues 

OwenCorrigan76:
When running Spotbugs against the project, it was throwing the following errors:

```[ERROR] Low: KafkaMetricsCollector.prefix not initialized in constructor and dereferenced in io.strimzi.kafka.metrics.KafkaMetricsCollector.metricName(MetricName) [io.strimzi.kafka.metrics.KafkaMetricsCollector] At KafkaMetricsCollector.java:[line 76] UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR
[ERROR] Low: KafkaPrometheusMetricsReporter.httpServer not initialized in constructor and dereferenced in io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter.getPort() [io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter] At KafkaPrometheusMetricsReporter.java:[line 90] UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR
[ERROR] Low: KafkaPrometheusMetricsReporter.kafkaMetricsCollector not initialized in constructor and dereferenced in io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter.contextChange(MetricsContext) [io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter] At KafkaPrometheusMetricsReporter.java:[line 86] UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR
[ERROR] Low: KafkaPrometheusMetricsReporter.kafkaMetricsCollector not initialized in constructor and dereferenced in io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter.metricChange(KafkaMetric) [io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter] At KafkaPrometheusMetricsReporter.java:[line 55] UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR
[ERROR] Low: KafkaPrometheusMetricsReporter.kafkaMetricsCollector not initialized in constructor and dereferenced in io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter.metricRemoval(KafkaMetric) [io.strimzi.kafka.metrics.KafkaPrometheusMetricsReporter] At KafkaPrometheusMetricsReporter.java:[line 61] UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR```

[...]

PR #31, opened 2024-07-19
Fix spotbugs errors

mimaison:
Put default values wherever possible, otherwise assert not null before using fields not initialized in constructors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -160,6 +160,12 @@
       <version>${slf4j.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-annotations</artifactId>
+      <version>${spotbugs.version}</version>
+      <scope>provided</scope>
+    </dependency>
 
     <dependency>
       <groupId>io.prometheus</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -160,12 +160,6 @@
       <version>${slf4j.version}</version>
       <scope>provided</scope>
     </dependency>
-    <dependency>
-      <groupId>com.github.spotbugs</groupId>
-      <artifactId>spotbugs-annotations</artifactId>
-      <version>${spotbugs.version}</version>
-      <scope>provided</scope>
-    </dependency>
 
     <dependency>
       <groupId>io.prometheus</groupId>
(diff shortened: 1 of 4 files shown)


scholzj (commented):
I'm not sure how much we want to keep this in sync with the rest of the Strimzi code ... in Strimzi, we would normally:
* Keep the annotation to silent the warning in places where not initializing the field in the reasonable choice
* I guess we would normally just use `if` instead of `Objects.requireNonNull`

But I'm not sure how much we want to keep it in sync given it is a different project mostly with different people contributing. So maybe we can just go with it.

Any thoughts @strimzi/maintainers?

Frawless:
> I'm not sure how much we want to keep this in sync with the rest of the Strimzi code ... in Strimzi, we would normally:
> 
>     * Keep the annotation to silent the warning in places where not initializing the field in the reasonable choice
> 
>     * I guess we would normally just use `if` instead of `Objects.requireNonNull`
> 
> 
> But I'm not sure how much we want to keep it in sync given it is a different project mostly with different people contributing. So maybe we can just go with it.
> 
> Any thoughts @strimzi/maintainers?

I don't have strong opinion, I am fine with the both options. 

Just wonder if depedency for annotations should be removed as it might get useful in the future...on the other hand guess it cause some troubles with `unused dependency check` ?

im-konge:
> I guess we would normally just use `if` instead of `Objects.requireNonNull`

Yeah I think that using the `if` would be maybe better, because this will anyway produce NPE when the collector will not be initialized, or? But will there be a situation when it will not be configured? Or instead of NPE, should it throw some reasonable exception with some information?

> Keep the annotation to silent the warning in places where not initializing the field in the reasonable choice

Yeah for example for the `httpServer` it looks reasonable to have the annotation there, as it will be anyway configured as part of the `configure` method.

I'm not that familiar with the code, but how much is this related to #20 ?

mimaison:
> Just wonder if depedency for annotations should be removed as it might get useful in the future

I'm not sure it makes sense keeping an unused dependency just in case we need it in the future. If we need it later, it's easy to add it back.

mimaison:
Ok, I'll follow what's done in other projects and keep the annotations. I added some details to each to explain why it's fine to suppress the warning.

scholzj (approved):
Thanks

im-konge (approved):
Thanks :)

Issue #34, opened 2024-07-23
Reconsider the lifecycle of the HttpServer


mimaison:
Currently we have logic to start the Prometheus HTTP server but no logic ever closes it. We should decide on the lifecycle we want for that server and see whether it makes sense to close it at some point.

mimaison:
Done in https://github.com/strimzi/metrics-reporter/pull/40

PR #40, opened 2024-08-20
Keep track of HTTPServer instances

mimaison:
An attempt at addressing #34 by keeping track of all HTTPServer instances started so they can be properly closed whenever possible. You can have multiple HTTPServer instances when several Kafka components are started in the same JVM, for example in Kafka Connect or Kafka Streams. 

If multiple components attempt to create servers on the same listener, the same server will be reused. When all using components are shutdown, the server can also be closed.

### DIFF
--- a/src/main/java/io/strimzi/kafka/metrics/HttpServers.java
+++ b/src/main/java/io/strimzi/kafka/metrics/HttpServers.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+package io.strimzi.kafka.metrics;
+
+import io.prometheus.metrics.exporter.httpserver.HTTPServer;
+import io.prometheus.metrics.model.registry.PrometheusRegistry;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Static class to keep track of all the HTTP servers started by all the Kafka components in a JVM.
+ */
+public class HttpServers {
+
+    private static final Map<Listener, ServerCounter> SERVERS = new HashMap<>();
+
+    /**
+     * Get or start a new HTTP server.
+     * @param listener The host and port
+     * @param registry The Prometheus registry to expose
+     * @return A ServerCounter instance
+     * @throws IOException if the HTTP server does not exist and cannot be started
+     */
+    public synchronized static ServerCounter get(Listener listener, PrometheusRegistry registry) throws IOException {
+        ServerCounter serverCounter = SERVERS.get(listener);
+        if (serverCounter == null) {
+            serverCounter = new ServerCounter(listener, registry);
+            SERVERS.put(listener, serverCounter);
+        }
+        serverCounter.count.incrementAndGet();
+        return serverCounter;
+    }
+
+    /**
+     * Release an HTTP server instance. If no other components hold this instance, it is closed.
+     * @param serverCounter The HTTP server instance to release
+     */
+    public synchronized static void release(ServerCounter serverCounter) {
+        if (serverCounter.close()) {
+            SERVERS.remove(serverCounter.listener);
+        }
+    }
+
+    /**
+     * Class used to keep track of the HTTP server started on a listener.
+     */
+    public static class ServerCounter {
+
+        private final AtomicInteger count;
+        private final HTTPServer server;
+        private final Listener listener;
+
+        private ServerCounter(Listener listener, PrometheusRegistry registry) throws IOException {
+            this.count = new AtomicInteger();
+            this.server = HTTPServer.builder()
+                    .hostname(listener.host)
+                    .port(listener.port)
+                    .registry(registry)
+                    .buildAndStart();
+            this.listener = listener;
+        }
+
+        /**
+         * The port this HTTP server instance is listening on. If the listener port is 0, this returns the actual port
+         * that is used.
+         * @return The port number
+         */
+        public int port() {
+            return server.getPort();
+        }
+
+        private synchronized boolean close() {
+            int remaining = count.decrementAndGet();
+            if (remaining == 0) {
+                server.close();
+                return true;
+            }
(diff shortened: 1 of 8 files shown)


src/main/java/io/strimzi/kafka/metrics/KafkaPrometheusMetricsReporter.java
```diff
@@ -57,7 +56,7 @@ public void configure(Map<String, ?> map) {
         config = new PrometheusMetricsReporterConfig(map, registry);
         collector = new KafkaMetricsCollector();
         // Add JVM metrics
-        JvmMetrics.builder().register(registry);
+        JvmMetrics.builder().register();
```
mimaison, line 59: The method without an argument is idempotent and registers the JVM metrics onto the default metrics registry, which is what we use at runtime. This allow multiple components to run this code without causing exceptions as the metrics are only added once.

It would be nice to pass a registry and still have the idempotent behavior if it's the default registry. I may open a PR on https://github.com/prometheus/client_java to do that. In the meantime, we have to use to method without arguments.

src/main/java/io/strimzi/kafka/metrics/HttpServers.java
```diff
@@ -0,0 +1,86 @@
...
+import io.prometheus.metrics.exporter.httpserver.HTTPServer;
+import io.prometheus.metrics.model.registry.PrometheusRegistry;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Static class to keep track of all the HTTP servers started by all the Kafka components in a JVM.
```
scholzj, line 16: What is static on it?

src/main/java/io/strimzi/kafka/metrics/HttpServers.java
```diff
@@ -0,0 +1,86 @@
...
+    private static final Map<Listener, ServerCounter> SERVERS = new HashMap<>();
+
+    /**
+     * Get or start a new HTTP server.
+     * @param listener The host and port
+     * @param registry The Prometheus registry to expose
+     * @return A ServerCounter instance
+     * @throws IOException if the HTTP server does not exist and cannot be started
+     */
+    public synchronized static ServerCounter get(Listener listener, PrometheusRegistry registry) throws IOException {
```
scholzj, line 29: Should it be named something like `getOrCreate` if it is not a simple getter?

scholzj (commented):
Some nits. LGTM otherwise.

src/main/java/io/strimzi/kafka/metrics/HttpServers.java
mimaison: Right, "static" is not the right term. I meant it as a utility. But I guess we can just say "Class to keep track ..."

src/main/java/io/strimzi/kafka/metrics/HttpServers.java
mimaison: Yes I've renamed it to `getOrCreate()`

scholzj (approved):
Thanks.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -140,80 +136,20 @@ public String toString() {
...
-                    .port(listener.port)
-                    .registry(registry)
-                    .buildAndStart();
-            LOG.info("HTTP server started on listener http://{}:{}", listener.host, httpServer.getPort());
-            return Optional.of(httpServer);
-        } catch (BindException be) {
-            LOG.info("HTTP server already started");
-            return Optional.empty();
+            HttpServers.ServerCounter server = HttpServers.getOrCreate(listener, registry);
+            LOG.info("HTTP server started on listener http://{}:{}", listener.host, server.port());
```
k-wall, line 148: The log line is misleading as the server isn't necessarily being started here.

src/main/java/io/strimzi/kafka/metrics/HttpServers.java
```diff
@@ -0,0 +1,86 @@
...
+        private final Listener listener;
+
+        private ServerCounter(Listener listener, PrometheusRegistry registry) throws IOException {
+            this.count = new AtomicInteger();
+            this.server = HTTPServer.builder()
+                    .hostname(listener.host)
+                    .port(listener.port)
+                    .registry(registry)
+                    .buildAndStart();
+            this.listener = listener;
```
k-wall, line 65: I'd add a log statement for starting the server.

src/main/java/io/strimzi/kafka/metrics/HttpServers.java
```diff
@@ -0,0 +1,86 @@
...
+         * @return The port number
+         */
+        public int port() {
+            return server.getPort();
+        }
+
+        private synchronized boolean close() {
+            int remaining = count.decrementAndGet();
+            if (remaining == 0) {
+                server.close();
```
k-wall, line 80: and log on shutdown

mimaison: Done. I checked that the port is returned (and valid) after `close()` is called.

src/main/java/io/strimzi/kafka/metrics/HttpServers.java
```diff
@@ -0,0 +1,86 @@
...
+    private static final Map<Listener, ServerCounter> SERVERS = new HashMap<>();
+
+    /**
+     * Get or create a new HTTP server if there isn't an existing instance for the specified listener.
+     * @param listener The host and port
+     * @param registry The Prometheus registry to expose
+     * @return A ServerCounter instance
+     * @throws IOException if the HTTP server does not exist and cannot be started
+     */
+    public synchronized static ServerCounter getOrCreate(Listener listener, PrometheusRegistry registry) throws IOException {
```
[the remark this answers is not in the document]
mimaison: Yeah the registry used to be hard coded to the default registry as it's what all components have to use. But because there isn't a mechanism to clear a registry, we allowed passing it in for tests.

k-wall (approved):
lgtm

PR #49, opened 2024-10-04
Separate logic for Kafka and Yammer metrics

### LINKED ISSUE #48: Using KafkaPrometheusMetricsReporter should only require kafka-clients
Currently when KafkaPrometheusMetricsReporter is loaded it tries to load classes that come from the kafka_2.13 and yammer libraries. The kafka_2.13 and yammer JARs should only be required when using the server side reporter YammerPrometheusMetricsReporter.

mimaison:
This fixes #48 

This PR separates the Kafka and Yammer specific metric logic. This ensures that when using`KafkaPrometheusMetricsReporter`, only classes present in kafka-clients are used. On the other hand `YammerPrometheusMetricsReporter` can use all dependencies present in the Kafka distribution as it's only used on the server side.


ppatierno (approved):
I haven't looked much at the code itself in the PR but tested with the Strimzi HTTP bridge where we found the issue. Everything seems to work fine now but the metrics reporter configuration should be documented in the bridge repo. I opened https://github.com/strimzi/strimzi-kafka-bridge/issues/932 for that.

k-wall:
@mimaison could you add a sentence or two to the PR description outlining at a high level how the problem is solved.  It should help reviewers understand the intent.

k-wall (approved):
The change made me wonder if there should be a separate Maven artefact one for the client use-case and one for the broker use case.   That'd keep the server side concerns completely out of the client's classpath.

I think what you have is good enough for the initial versions.

Issue #55, opened 2024-10-25
Add support for dynamic reconfiguration


mimaison:
When running on the broker side, `MetricsReporter` plugins can be reconfigured at runtime. Currently metrics-reporter does not define any configurations as reconfigurable.

We should make the `prometheus.metrics.reporter.allowlist` configuration reconfigurable. It is often useful for operators to adjust the metrics collected at runtime in order to diagnose and investigate issues. This is currently possible when using jmx_reporter. 

None of the other metrics-reporter configurations really benefit from being reconfigured, so I propose to not update them.

mimaison:
This was implemented in https://github.com/strimzi/metrics-reporter/pull/73

PR #73, opened 2025-02-27
Split the reporter into client and server modules

mimaison:
This implements [Proposal 96](https://github.com/strimzi/proposals/blob/main/096-split-metrics-reporter-into-modules.md) and splits the reporter into 2 modules:
- client-metrics-reporter
- server-metrics-reporter

This also adds support for dynamic reconfiguration (#55).

I'm still unsure about the packaging. Running `mvn package` produces archives under client-metrics-reporter and server-metrics-reporter. Is that sufficient? or do we want a single archive?


mimaison:
The CI is failing. I can reproduce locally but not sure I understand why this is happening.

Running:
```
mvn verify
mvn spotbugs:check
```
Hits the issue:
```
[ERROR] Failed to execute goal on project server-metrics-reporter: Could not resolve dependencies for project io.strimzi:server-metrics-reporter:jar:1.0.0-SNAPSHOT
[ERROR] dependency: io.strimzi:client-metrics-reporter:jar:1.0.0-SNAPSHOT (compile)
[ERROR] 	Could not find artifact io.strimzi:client-metrics-reporter:jar:1.0.0-SNAPSHOT
[ERROR] dependency: io.strimzi:client-metrics-reporter:jar:tests:1.0.0-SNAPSHOT (test)
[ERROR] 	Could not find artifact io.strimzi:client-metrics-reporter:jar:tests:1.0.0-SNAPSHOT
[ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project server-metrics-reporter: Could not resolve dependencies for project io.strimzi:server-metrics-reporter:jar:1.0.0-SNAPSHOT
dependency: io.strimzi:client-metrics-reporter:jar:1.0.0-SNAPSHOT (compile)
	Could not find artifact io.strimzi:client-metrics-reporter:jar:1.0.0-SNAPSHOT
dependency: io.strimzi:client-metrics-reporter:jar:tests:1.0.0-SNAPSHOT (test)
	Could not find artifact io.strimzi:client-metrics-reporter:jar:tests:1.0.0-SNAPSHOT
```

But if I run:
```
mvn verify spotbugs:check
```
then it works

README.md
```diff
@@ -37,47 +37,51 @@ The metrics reporter has the following configurations:
 ### Kafka Brokers
```
k-wall, line 37: Should this say Kafka Brokers and Controllers?

see-quick:
> The CI is failing. I can reproduce locally but not sure I understand why this is happening.
> 
> Running:
> 
> ```
> mvn verify
> mvn spotbugs:check
> ```
> 
> Hits the issue:
> 
> ```
> [ERROR] Failed to execute goal on project server-metrics-reporter: Could not resolve dependencies for project io.strimzi:server-metrics-reporter:jar:1.0.0-SNAPSHOT
> [ERROR] dependency: io.strimzi:client-metrics-reporter:jar:1.0.0-SNAPSHOT (compile)
> [ERROR] 	Could not find artifact io.strimzi:client-metrics-reporter:jar:1.0.0-SNAPSHOT
> [ERROR] dependency: io.strimzi:client-metrics-reporter:jar:tests:1.0.0-SNAPSHOT (test)
> [ERROR] 	Could not find artifact io.strimzi:client-metrics-reporter:jar:tests:1.0.0-SNAPSHOT
> [ERROR] -> [Help 1]
> org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project server-metrics-reporter: Could not resolve dependencies for project io.strimzi:server-metrics-reporter:jar:1.0.0-SNAPSHOT
> dependency: io.strimzi:client-metrics-reporter:jar:1.0.0-SNAPSHOT (compile)
> 	Could not find artifact io.strimzi:client-metrics-reporter:jar:1.0.0-SNAPSHOT
> dependency: io.strimzi:client-metrics-reporter:jar:tests:1.0.0-SNAPSHOT (test)
> 	Could not find artifact io.strimzi:client-metrics-reporter:jar:tests:1.0.0-SNAPSHOT
> ```
> 
> But if I run:
> 
> ```
> mvn verify spotbugs:check
> ```
> 
> then it works

I think `mvn verify spotbugs:check` works because during the verify phase, it shares artifacts in memory but if you try to run:
1. mvn verify 
2. mvn spotbugs:check

then verify does not install artifacts to local repo `.m2` and that those dependencies are not persistent for the next run of the mvn command (i.e., mvn spot bugs:check). To make those dependencies persistent, you should use `mvn install -DskipTests` I think.

client-metrics-reporter/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+    <parent>
+        <groupId>io.strimzi</groupId>
+        <artifactId>metrics-reporter</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>client-metrics-reporter</artifactId>
+
+    <dependencies>
+        <dependency>
```
k-wall, line 15: For multi module projects, maven best practice is to express shared dependencies in the parent POM using a dependencyManagement section.  You can then remove the version numbers from the dependencies in the child poms.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
```diff
@@ -98,17 +105,26 @@ public void validateReconfiguration(Map<String, ?> configs) throws ConfigExcepti
...
     public Set<String> reconfigurableConfigs() {
-        return Collections.emptySet();
+        return Set.of();
     }
 
     @Override
     public void contextChange(MetricsContext metricsContext) {
         String prefix = metricsContext.contextLabels().get(MetricsContext.NAMESPACE);
+        if (!PREFIXES.contains(prefix)) {
+            throw new IllegalStateException("ClientMetricsReporter should only be used in Kafka servers");
```
k-wall, line 115: ClientMetricsReporter should only be used by clients?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/common/AbstractReporter.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Common reporter logic to track metrics that match an allowlist pattern.
+ */
+public abstract class AbstractReporter {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractReporter.class);
+
+    private final Map<Object, MetricWrapper> allowedMetrics = new ConcurrentHashMap<>();
```
k-wall, line 23: I don't find the naming of `otherMetrics` too clear.   It took me a while to grok what was going on:

You keep track of the otherMetrics so that in the case the allow list is changed, you can start to report on the newly matched metrics.  You only need to do this if the implementation is reconfigurable.

Can I ask for a few more comments (class level)?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/common/AbstractReporter.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public Collection<MetricWrapper> allowedMetrics() {
+        return allowedMetrics.values();
+    }
+
+    /**
+     * Update the allowed metrics based on the current allowlist pattern.
+     */
+    public void updateAllowedMetrics() {
+        if (!isReconfigurable()) return;
```
k-wall, line 75: We are guaranteed that this method is called by a single thread, right?

README.md
mimaison: I updated that section. I think we'll probably want further improvements to the README, it's still a bit bare bones.

client-metrics-reporter/pom.xml
mimaison: Thanks for the tip! I already had the versions defined in the parent POM so I removed the versions from the modules.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
mimaison: Yep, fixed.

mimaison:
> I think `mvn verify spotbugs:check` works because during the verify phase, it shares artifacts in memory but if you try to run:
> 
>     1. mvn verify
> 
>     2. mvn spotbugs:check
> 
> 
> then verify does not install artifacts to local repo `.m2` and that those dependencies are not persistent for the next run of the mvn command (i.e., mvn spot bugs:check). To make those dependencies persistent, you should use `mvn install -DskipTests` I think.

So what do you recommend? Should I add a step to run `mvn install`, or run `verify` and `spotbugs:check` together? What do other Strimzi projects do?

ppatierno:
I was looking at the bridge as simpler compared to the operator. We have the verify and spotbugs as two separated stages anyway, you can see here:

https://github.com/strimzi/strimzi-kafka-bridge/blob/main/.azure/templates/jobs/build_java.yaml#L27

where the `java_verify` is a make target running mvn verify.
and then here:

https://github.com/strimzi/strimzi-kafka-bridge/blob/main/.azure/templates/jobs/build_java.yaml#L35

where the `spotbugs` is a make target running mvn spotbugs:check.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/common/AbstractReporter.java
ppatierno: I agree with Keith, it would be better to have some comments.

mimaison: I renamed `otherMetrics` to `disallowedMetrics`. I also added a few comments to this class to clarify its role. Let me know if some parts are still not clear.

mimaison:
> I was looking at the bridge as simpler compared to the operator. We have the verify and spotbugs as two separated stages anyway, you can see here:
> 
> https://github.com/strimzi/strimzi-kafka-bridge/blob/main/.azure/templates/jobs/build_java.yaml#L27
> 
> where the `java_verify` is a make target running mvn verify. and then here:
> 
> https://github.com/strimzi/strimzi-kafka-bridge/blob/main/.azure/templates/jobs/build_java.yaml#L35
> 
> where the `spotbugs` is a make target running mvn spotbugs:check.

I'm not sure I follow. What do you suggest doing?

see-quick:
> > I think `mvn verify spotbugs:check` works because during the verify phase, it shares artifacts in memory but if you try to run:
> > ```
> > 1. mvn verify
> > 
> > 2. mvn spotbugs:check
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > then verify does not install artifacts to local repo `.m2` and that those dependencies are not persistent for the next run of the mvn command (i.e., mvn spot bugs:check). To make those dependencies persistent, you should use `mvn install -DskipTests` I think.
> 
> So what do you recommend? Should I add a step to run `mvn install`, or run `verify` and `spotbugs:check` together? What do other Strimzi projects do?

Well we simply do 
```
      # Build the Java code without tests
      - bash: "make java_install"
        displayName: "Build Java code"
        env:
          MVN_ARGS: "-DskipTests -e -V -B"
          CURL_ARGS: "-s"

      # Run different checks on the code
      - bash: "make spotbugs"
        displayName: "Run Spotbugs"
        env:
          MVN_ARGS: "-e -V -B"
```
where make `java_install`, which will trigger `java_build` and then it would trigger `./build.sh` and inside `build.sh` where in makefile we define
```
.PHONY: build clean release all $(SUBDIRS) $(DOCKER_TARGETS) spotbugs
```
so Spotbugs is part of the building process (but last to execute). I would go with mvn verify + spotbugs in one step IMHO.

mimaison:
I merged both steps but it's not triggering the CI now

k-wall:
lgtm - thanks @mimaison

ppatierno (approved):
LGTM!

mimaison:
Thanks for the approvals.

Nobody commented on my query regarding the packaging:
> I'm still unsure about the packaging. Running mvn package produces archives under client-metrics-reporter and server-metrics-reporter. Is that sufficient? or do we want a single archive?

see-quick (approved):

k-wall:
You are talking about client-metrics-reporter/target/client-metrics-reporter-1.0.0-SNAPSHOT.tar.gz and server-metrics-reporter/target/server-metrics-reporter-1.0.0-SNAPSHOT.tar.gz right?

Separate archives seems right to me (after all they have separate use-cases), but I'm not sufficient close to Strimzi to know if this matches its norms.

k-wall:
I'm actually seeing a javadoc error from the build, which is fatal.


```
mvn --version
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: /Users/kwall/.sdkman/candidates/maven/current
Java version: 21, vendor: Eclipse Adoptium, runtime: /Users/kwall/.sdkman/candidates/java/21-tem
Default locale: en_GB, platform encoding: UTF-8
OS name: "mac os x", version: "13.7.4", arch: "x86_64", family: "mac"
```


```
[INFO] No previous run data found, generating javadoc.
[WARNING] Javadoc Warnings
[WARNING] /Users/kwall/src/metrics-reporter/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/common/AbstractReporter.java:20: warning: use of default constructor, which does not provide a comment
[WARNING] public abstract class AbstractReporter {
[WARNING] ^
[WARNING] /Users/kwall/src/metrics-reporter/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/common/DataPointSnapshotBuilder.java:20: warning: use of default constructor, which does not provide a comment
[WARNING] public class DataPointSnapshotBuilder {
[WARNING] ^
[WARNING] /Users/kwall/src/metrics-reporter/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/HttpServers.java:20: warning: use of default constructor, which does not provide a comment
[WARNING] public class HttpServers {
[WARNING] ^
[WARNING] /Users/kwall/src/metrics-reporter/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java:76: warning: use of default constructor, which does not provide a comment
[WARNING] public static class ListenerValidator implements ConfigDef.Validator {
[WARNING] ^
[WARNING] 4 warnings
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.6.3:jar (attach-javadocs) on project client-metrics-reporter: MavenReportException: Error while generating Javadoc: Project contains Javadoc Warnings -> [Help 1]
```

mimaison:
The CI did not hit these errors. What command did you run?

k-wall:
> The CI did not hit these errors. What command did you run?

`mvn package`

fvaleri:
`mvn package` works fine with Maven 3.9.6

mimaison:
Yeah I can reproduce with Java 21, but it works fine with 17.

k-wall (approved):

mimaison:
I think it would be good to fix the javadoc warnings, but we can do it in another PR. I opened https://github.com/strimzi/metrics-reporter/issues/77.

ppatierno:
> Separate archives seems right to me (after all they have separate use-cases), but I'm not sufficient close to Strimzi to know if this matches its norms.

I think it's first time for Strimzi dealing with something like this.
Assuming we have 2 maven packages anyway ... having a single release artifact means that the two modules will be released together so if you need to update just clients, for example, and release it, the artifact will contain a new version of server even without changes in it (and vice versa). Multiple release artifacts give the flexibility to the modules to evolve separately but I think it could be more complex to deal with a double release process, possibility of dependencies conflicts if the two modules evolve by using different versions of same dependencies, or?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/common/AbstractReporter.java
```diff
@@ -0,0 +1,105 @@
...
+     * @return A collection of MetricWrapper
+     */
+    public Collection<MetricWrapper> allowedMetrics() {
+        return allowedMetrics.values();
+    }
+
+    /**
+     * Update the allowed metrics based on the current allowlist pattern.
+     */
+    public void updateAllowedMetrics() {
```
fvaleri, line 86: This is slightly inefficient. You can use iterators and avoid creating a temporary map, which is also faster.

scholzj:
> You are talking about client-metrics-reporter/target/client-metrics-reporter-1.0.0-SNAPSHOT.tar.gz and server-metrics-reporter/target/server-metrics-reporter-1.0.0-SNAPSHOT.tar.gz right?
> 
> Separate archives seems right to me (after all they have separate use-cases), but I'm not sufficient close to Strimzi to know if this matches its norms.

I guess the point of the archives is that you can drop them into the Kafka binaries for the server components. You do not want to do it twice. So, the question is if Brokers/Connect/etc. can do only with the server archive or not. If copying there the server archive is sufficient, then I think it is fine to have it split.

scholzj:
> The current code generates 2 sets of archives:
> 
> * `client-metrics-reporter/target/client-metrics-reporter-1.0.0-SNAPSHOT.tar.gz` and `.zip`
> * `server-metrics-reporter/target/server-metrics-reporter-1.0.0-SNAPSHOT.tar.gz` and `.zip`
> 
> If you want to use the reporter with your clients, you add the JARs from the client archive to the classpath.
> 
> Similarly if you want to use the reporter with your brokers/controllers, you add the JARs from the server archive to the classpath. (The server archive contains the client JARs)

And what about things such as Connect ... do they also need only the server one?

mimaison:
> And what about things such as Connect ... do they also need only the server one?

Connect and Streams are clients, they need the client-metrics-reporter JARs.

scholzj:
> > And what about things such as Connect ... do they also need only the server one?
> 
> Connect and Streams are clients, they need the client-metrics-reporter JARs.

I would say that we should have a single archive that users unpack into the Apache Kafka distribution to cover both Kafka controllers, brokers, Connect and MM2. Otherwise this would be confusing what needs which archive and likely duplicating dependencies as well.

ppatierno:
+1 for one archive from me as well.

@scholzj when you say "users unpack" are you referring to the use case when they want to use the metrics reporter outside of Strimzi right?
Because with Strimzi we are going to put such an archive into the Kafka base image already, or?

scholzj:
> +1 for one archive from me as well.
> 
> @scholzj when you say "users unpack" are you referring to the use case when they want to use the metrics reporter outside of Strimzi right? Because with Strimzi we are going to put such an archive into the Kafka base image already, or?

Yes, that would be outside of Strimzi. In Strimzi we would pull it into the image through 3rd-party libs - at least that would be my expectation.

mimaison:
I've pushed changes to generate a single set of archives for the release artifacts. They are named `metrics-reporter-<version>` and are put under `/target`. 

We still need to generate a package for with only the client side library for the client integration tests. This is why there are 2 `dist.xml` file:
- client-metrics-reporter: it generates a folder under `client-metrics-reporter/target` for the client integration test
- server-metrics-reporter: it generates the release artifacts

client-metrics-reporter/src/main/assembly/dist.xml
```diff
@@ -0,0 +1,19 @@
+<!-- Dist for generating a package only containing client libraries for client integration tests-->
+<assembly xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2 http://maven.apache.org/xsd/assembly-1.1.2.xsd">
+    <id>zip</id>
+    <includeBaseDirectory>true</includeBaseDirectory>
+
+    <formats>
+        <format>dir</format>
```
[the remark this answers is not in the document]
mimaison: I have to admit I'm not sure what I'm doing here. Let me try explaining what I'm thinking.

In terms of release artifacts, we want client and server side JARs (that we push to Maven repositories), and archives containing both client and server JARs + all the dependencies that we attach to our GitHub release.

If I understand correctly, JARs are produced by `maven-jar-plugin` and we use `maven-assembly-plugin` to produce the archives. The archives we want are basically the output from `server-metrics-reporter`. 

Users wanting to use the reporter in their client side applications can add the `io.strimzi.client-metrics-reporter` dependency to their project, or unzip the release archive and add them to their classpath. From Connect or Kafka clusters, I think users will typically unzip the release archive and add them to their classpath or to the Kafka `libs` directory. Note that for client applications and Connect, if you add the archives to your classpath, you'll have `server-metrics-reporter-<VERSION>.jar` in the classpath which will be unused, as these only need the client JAR. We discussed above whether to have different archives for client and server but people seem to prefer only have one.

The reason I added `maven-assembly-plugin` to `client-metrics-reporter` is to produce a directory with the client JARs and all the dependencies that can be used for integration tests. We don't want this to be a release artifact.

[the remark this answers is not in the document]
mimaison: cc @im-konge as you are the last person to run a release, does the above make sense?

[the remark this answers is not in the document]
k-wall: @mimaison thanks for the explanation, yes it made sense.

mimaison:
Two maintainers have approved the PR, I'll wait a couple of days and if there isn't new comments I'll merge it.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/common/AbstractReporter.java
fvaleri: What about this comment?

mimaison: Can you clarify what sort of logic you have in mind?

fvaleri: > Can you clarify what sort of logic you have in mind?

Exactly the same logic, but using iterators.

In addition to save an extra objects allocation, you also save an extra loop at the end caused by putAll. I think this kind of little optimizations matter when dealing with a lot of metrics.

```sh
    /**
     * Update the allowed metrics based on the current allowlist pattern.
     */
    public void updateAllowedMetrics() {
        if (!isReconfigurable()) return;

        Iterator<Map.Entry<Object, MetricWrapper>> disallowedIt = disallowedMetrics.entrySet().iterator();
        while (disallowedIt.hasNext()) {
            Map.Entry<Object, MetricWrapper> entry = disallowedIt.next();
            if (matches(entry.getValue().prometheusName())) {
                allowedMetrics.put(entry.getKey(), entry.getValue());
                disallowedIt.remove();
            }
        }

        Iterator<Map.Entry<Object, MetricWrapper>> allowedIt = allowedMetrics.entrySet().iterator();
        while (allowedIt.hasNext()) {
            Map.Entry<Object, MetricWrapper> entry = allowedIt.next();
            if (!matches(entry.getValue().prometheusName())) {
                disallowedMetrics.put(entry.getKey(), entry.getValue());
                allowedIt.remove();
            }
        }
    }
```

mimaison: This logic is not on the metrics collection path. It is only invoked when the allowlist is updated at runtime on the server side.

I'm leaning towards keeping the code as is for now. If we find it inefficient we can try optimizing it in the future.

fvaleri: I agree that this is not on a critical path, but disallowedMetrics and allowedMetrics contains all the metrics, which could be a lot. Anyway, up to you to decide.

PR #74, opened 2025-02-27
Fix spotbugs errors and CI

fvaleri:
Currently the spotbugs check on the CI is executed before project build, so it's a noop. This is the reason why we have a bunch of spotbugs errors that I'm also fixing here.

```sh
[INFO] BugInstance size is 8
[INFO] Error size is 0
[INFO] Total bugs: 8
[ERROR] Medium: io.strimzi.kafka.metrics.MetricWrapper.labels() may expose internal representation by returning MetricWrapper.labels [io.strimzi.kafka.metrics.MetricWrapper] At MetricWrapper.java:[line 39] EI_EXPOSE_REP
[ERROR] Medium: Public static io.strimzi.kafka.metrics.PrometheusCollector.register(PrometheusRegistry) may expose internal representation by returning PrometheusCollector.INSTANCE [io.strimzi.kafka.metrics.PrometheusCollector] At PrometheusCollector.java:[line 42] MS_EXPOSE_REP
[ERROR] Medium: Exception thrown in class io.strimzi.kafka.metrics.PrometheusMetricsReporterConfig at new io.strimzi.kafka.metrics.PrometheusMetricsReporterConfig(Map, PrometheusRegistry) will leave the constructor. The object under construction remains partially initialized and may be vulnerable to Finalizer attacks. [io.strimzi.kafka.metrics.PrometheusMetricsReporterConfig, io.strimzi.kafka.metrics.PrometheusMetricsReporterConfig] At PrometheusMetricsReporterConfig.java:[line 84]At PrometheusMetricsReporterConfig.java:[line 84] CT_CONSTRUCTOR_THROW
[ERROR] Medium: new io.strimzi.kafka.metrics.PrometheusMetricsReporterConfig(Map, PrometheusRegistry) may expose internal representation by storing an externally mutable object into PrometheusMetricsReporterConfig.registry [io.strimzi.kafka.metrics.PrometheusMetricsReporterConfig] At PrometheusMetricsReporterConfig.java:[line 86] EI_EXPOSE_REP2
[ERROR] Medium: Exception thrown in class io.strimzi.kafka.metrics.http.HttpServers$ServerCounter at new io.strimzi.kafka.metrics.http.HttpServers$ServerCounter(Listener, PrometheusRegistry) will leave the constructor. The object under construction remains partially initialized and may be vulnerable to Finalizer attacks. [io.strimzi.kafka.metrics.http.HttpServers$ServerCounter, io.strimzi.kafka.metrics.http.HttpServers$ServerCounter] At HttpServers.java:[line 69]At HttpServers.java:[line 69] CT_CONSTRUCTOR_THROW
[ERROR] Low: Overridable method addCollector is called from constructor new io.strimzi.kafka.metrics.kafka.KafkaCollector(PrometheusCollector). [io.strimzi.kafka.metrics.kafka.KafkaCollector] At KafkaCollector.java:[line 45] MC_OVERRIDABLE_METHOD_CALL_IN_CONSTRUCTOR
[ERROR] Medium: Public static io.strimzi.kafka.metrics.kafka.KafkaCollector.getCollector(PrometheusCollector) may expose internal representation by returning KafkaCollector.INSTANCE [io.strimzi.kafka.metrics.kafka.KafkaCollector] At KafkaCollector.java:[line 58] MS_EXPOSE_REP
[ERROR] Medium: Public static io.strimzi.kafka.metrics.yammer.YammerCollector.getCollector(PrometheusCollector) may expose internal representation by returning YammerCollector.INSTANCE [io.strimzi.kafka.metrics.yammer.YammerCollector] At YammerCollector.java:[line 62] MS_EXPOSE_REP
```

### DIFF
--- a/.azure/templates/jobs/build_java.yaml
+++ b/.azure/templates/jobs/build_java.yaml
@@ -28,10 +28,6 @@ jobs:
       - template: '../steps/prerequisites/install_java.yaml'
         parameters:
           JDK_VERSION: $(jdk_version)
-      - bash: "mvn ${MVN_ARGS} spotbugs:check"
-        displayName: "Spotbugs"
-        env:
-          MVN_ARGS: "-e -V -B"
       - bash: "mvn ${MVN_ARGS} verify"
         displayName: "Build & Test Java"
         env:
@@ -39,6 +35,10 @@ jobs:
           BRANCH: $(Build.SourceBranch)
           TESTCONTAINERS_RYUK_DISABLED: "TRUE"
           TESTCONTAINERS_CHECKS_DISABLE: "TRUE"
+          MVN_ARGS: "-e -V -B"
+      - bash: "mvn ${MVN_ARGS} spotbugs:check"
+        displayName: "Spotbugs"
+        env:
           MVN_ARGS: "-e -V -B"
       # We have to TAR the target directory to maintain the permissions of
       # the files which would otherwise change when downloading the artifact
--- a/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
+++ b/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
@@ -13,7 +13,6 @@ import org.apache.kafka.common.config.ConfigException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -24,7 +23,7 @@ import java.util.stream.Collectors;
 /**
 * Configuration for the PrometheusMetricsReporter implementation.
 */
-public class PrometheusMetricsReporterConfig extends AbstractConfig {
+public final class PrometheusMetricsReporterConfig extends AbstractConfig {
 
     private static final Logger LOG = LoggerFactory.getLogger(PrometheusMetricsReporterConfig.class);
     private static final String CONFIG_PREFIX = "prometheus.metrics.reporter.";
@@ -145,13 +144,8 @@ public class PrometheusMetricsReporterConfig extends AbstractConfig {
             LOG.info("HTTP server listener not enabled");
             return Optional.empty();
         }
-        try {
-            HttpServers.ServerCounter server = HttpServers.getOrCreate(listener, registry);
-            LOG.info("HTTP server listening on http://{}:{}", listener.host, server.port());
-            return Optional.of(server);
-        } catch (IOException ioe) {
-            LOG.error("Failed starting HTTP server", ioe);
-            throw new RuntimeException(ioe);
-        }
+        HttpServers.ServerCounter server = HttpServers.getOrCreate(listener, registry);
+        LOG.info("HTTP server listening on http://{}:{}", listener.host, server.port());
+        return Optional.of(server);
     }
 }
--- a/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
+++ b/src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
@@ -4,6 +4,7 @@
  */
 package io.strimzi.kafka.metrics;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.prometheus.metrics.model.registry.PrometheusRegistry;
 import io.strimzi.kafka.metrics.http.HttpServers;
 import io.strimzi.kafka.metrics.http.Listener;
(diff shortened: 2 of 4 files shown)


.azure/templates/jobs/build_java.yaml
```diff
@@ -53,4 +53,4 @@ jobs:
           testResultsFormat: JUnit
           testResultsFiles: '**/TEST-*.xml'
           testRunTitle: "Unit & Integration tests"
-        condition: always()
```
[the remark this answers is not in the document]
fvaleri: That's strange, as I have these newlines locally.

src/main/java/io/strimzi/kafka/metrics/PrometheusMetricsReporterConfig.java
```diff
@@ -24,7 +23,7 @@
 /**
 * Configuration for the PrometheusMetricsReporter implementation.
 */
-public class PrometheusMetricsReporterConfig extends AbstractConfig {
+public final class PrometheusMetricsReporterConfig extends AbstractConfig {
```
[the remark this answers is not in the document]
fvaleri: Forgot to remove from a previous test. Sorry.

.azure/templates/jobs/build_java.yaml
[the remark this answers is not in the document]
mimaison: My bad, your PR added the new lines, so all good.

fvaleri:
Wait, there was indeed a reason for that final. Let me see if I can fix in another way.

fvaleri:
@mimaison this is a quick fix, otherwise we would need a major refactoring of `PrometheusMetricsReporterConfig`, for example using the builder pattern, but I think it is not desired at this stage.

ppatierno (approved):
The fixes look reasonable to me.

see-quick (commented):
I wonder if this might be a copy-paste error? Maybe the issue originates from where it was copied?  @mimaison?

fvaleri:
> I wonder if this might be a copy-paste error? Maybe the issue originates from where it was copied? @mimaison?

Good point. Maybe there is some wrong template somewhere, as the same configuration error was present in the HTTP Bridge. I guess it is worth checking, but it is not related to this PR.

src/main/java/io/strimzi/kafka/metrics/http/HttpServers.java
```diff
@@ -53,22 +53,33 @@ public synchronized static void release(ServerCounter serverCounter) {
      * Class used to keep track of the HTTP server started on a listener.
      */
     public static class ServerCounter {
+        private HTTPServer.Builder builder;
+        private Listener listener;
+        private AtomicInteger count;
```
mimaison, line 58: Why are we making these fields not `final`?

src/main/java/io/strimzi/kafka/metrics/http/HttpServers.java
```diff
@@ -53,22 +53,33 @@ public synchronized static void release(ServerCounter serverCounter) {
...
-            this.server = builder.buildAndStart();
-            LOG.debug("Started HTTP server on http://{}:{}", listener.host, server.getPort());
             this.listener = listener;
+            this.count = new AtomicInteger();
+        }
+
+        /**
+         * Start the HTTP server.
+         */
+        public void start() {
```
mimaison, line 75: In `PrometheusMetricsReporterConfig` you used the `@SuppressFBWarnings("CT_CONSTRUCTOR_THROW")` annotation to silence this Spotbug warning, but here you decided to wrap `IOException` into an unchecked exception instead.
Can you explain why the 2 different approaches?
Also if you decide to keep the `start()` method, it should be private.

fvaleri: You can't throw an exception from the constructor, that what needs to change to avoid the finalizer attack.

> Can you explain why the 2 different approaches?

To me it seems natural to create and then start a server (two separate operations), while the allow list compilation seems more related to initialization. For `PrometheusMetricsReporterConfig`, a good alternative could be to lazy compile the allow list on the very first call to `isAllowed`. Wdyt?

> Also if you decide to keep the start() method, it should be private.

Done.

src/main/java/io/strimzi/kafka/metrics/http/HttpServers.java
fvaleri: Fixed.

src/main/java/io/strimzi/kafka/metrics/http/HttpServers.java
mimaison: > For PrometheusMetricsReporterConfig, a good alternative could be to lazy compile the allow list on the very first call to isAllowed. Wdyt?

I prefer that the reporter fails to startup if a bad allowlist is provided instead of failing at some random point in the future when the `/metrics` endpoint is scrapped.

mimaison: > To me it seems natural to create and then start a server (two separate operations), while the allow list compilation seems more related to initialization. 

Fair enough, but why are we wrapping `IOException` into an unchecked exception if it's not thrown by the constructor anymore?

mimaison (approved):
LGTM

see-quick (approved):

PR #81, opened 2025-04-10
Add integration test for Connect metrics

### LINKED ISSUE #63: Add integration test for Connect
We should add an integration test that check metrics are correctly emitted when using the metrics reporter with Connect.

mimaison:
Fixes #63

### DIFF
--- a/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
+++ b/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
@@ -24,11 +24,12 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * Utility class to create and retrieve metrics
@@ -46,9 +47,8 @@ public class MetricsUtils {
      * Query the HTTP endpoint and returns the output
      * @param port The port to query
      * @return The lines from the output
-     * @throws Exception If any error occurs
      */
-    public static List<String> getMetrics(int port) throws Exception {
+    public static List<String> getMetrics(int port) {
         return getMetrics("localhost", port);
     }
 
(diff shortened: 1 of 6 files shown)


client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/integration/TestConnectMetricsIT.java
```diff
@@ -0,0 +1,275 @@
...
+            for (int i = 0; i < 5; i++) {
+                producer.send(new ProducerRecord<>(TOPIC, "record" + i));
+            }
+        }
+    }
+
+    @AfterEach
+    public void tearDown() {
+        connect.stop();
+        kafka.stop();
```
see-quick, line 137: Add null check here :) and if it's `!= null` then invoke stop() to prevent potential NPE.

see-quick (commented):
Looks good, just a nit.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
```diff
@@ -104,32 +110,31 @@ public static void assertInfoSnapshot(MetricSnapshot snapshot, Labels labels, St
...
+            for (Pattern pattern : expectedPatterns) {
+                while (true) {
+                    try {
+                        List<String> filteredMetrics = filterMetrics(metrics, pattern);
+                        condition.accept(filteredMetrics);
+                        break;
+                    } catch (Throwable t) {
+                        assertInstanceOf(AssertionError.class, t);
+                        TimeUnit.MILLISECONDS.sleep(100L);
+                        metrics = getMetrics(container.getHost(), container.getMappedPort(port));
```
katheris, line 136: I know this was here before, but I'm struggling to follow how this works. As far as I understand `condition.accept()` will throw an `AssertionError` if the condition fails. So `assertInstanceOf` is asserting the throwable is the correct type. Assuming it is the right type, we then sleep and reset the metrics to run the while loop again.

Why aren't we just letting the `AssertionError` be thrown and fail the test immediately? What is the sleep for? Also why are we resetting the `metrics` variable? Finally why is the `while (true)` needed?

katheris (commented):
The tests look fine, but some of the code in the MetricsUtils class I'm struggling to follow

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
mimaison: > Why aren't we just letting the AssertionError be thrown and fail the test immediately? What is the sleep for? Also why are we resetting the metrics variable? Finally why is the while (true) needed?

The reason is that in some cases metrics take some time to appear. So this basically retries for up to `TIMEOUT` for metrics to be there. This is mostly seen with Connect where connectors are started asynchronously so metrics might not be there immediately.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
```diff
@@ -55,22 +57,26 @@ public static List<String> getMetrics(int port) throws Exception {
...
-        URL url = new URL("http://" + host + ":" + port + "/metrics");
-        HttpURLConnection con = (HttpURLConnection) url.openConnection();
-        con.setRequestMethod("GET");
-
-        try (BufferedReader in = new BufferedReader(new InputStreamReader(con.getInputStream()))) {
-            String inputLine;
-            while ((inputLine = in.readLine()) != null) {
-                if (!inputLine.startsWith("#")) {
-                    metrics.add(inputLine);
+        assertTimeoutPreemptively(TIMEOUT, () -> {
```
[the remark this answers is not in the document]
mimaison: I've seen the `GET` call fail if it's queried too early, hence why I wrapped it in `assertTimeoutPreemptively()` so it's retried.
This is called from multiple places and yes in cases it comes from `verify()` that's 2 retriable blocks.

mimaison:
@katheris Have I addressed your comments?

katheris (approved):
Thanks for the explanations to my questions, all makes sense so happy to approve.

PR #97, opened 2025-06-18
Bump Kafka to 4.0.0

mimaison:


see-quick (approved):
LGTM 👍

mimaison:
Actually this does not work with Java 11. The server side reporter depends on `org.apache.kafka/kafka_2.13` which does not support Java 11 anymore.
So we'd need to run the `client-metrics-reporter` tests with both 11 and 17 and the `server-metrics-reporter` tests with only 17.

scholzj (approved):
LGTM ... assuming it works 😉

PR #114, opened 2026-01-28
Add RELEASING.md

mimaison:
Document the release process.


RELEASING.md
```diff
@@ -0,0 +1,53 @@
...
+## Run the Release GitHub Action
+
+Run the [Release](https://github.com/strimzi/metrics-reporter/actions/workflows/release.yml) Action. Set the branch and tag for the release.
+Once complete, notify a Strimzi Maintainer to publish the artifacts to Maven.
+
+## Update the GitHub Milestone
+
+Close the current GitHub [milestone](https://github.com/strimzi/metrics-reporter/milestones) and create the next one.
+
+## Create the GitHub Release
```
im-konge: Should we mention that the zip and tar should be attached to the release?

And as an example, we can add https://repo1.maven.org/maven2/io/strimzi/server-metrics-reporter/0.3.0-rc1/server-metrics-reporter-0.3.0-rc1.tar.gz and https://repo1.maven.org/maven2/io/strimzi/server-metrics-reporter/0.3.0-rc1/server-metrics-reporter-0.3.0-rc1.zip

WDYT?

im-konge (commented):
Just one comment. Do you want to update it with the GA info after the 0.3.0 release? (It's the same thing, you will again tag it, push the tag, create release announcement, where you set it as the latest release, and then announce it everywhere as the RC).

And thanks for writing this :)

katheris (commented):
@mimaison thanks for writing this, just as a note for Access Operator we made the release notes match the Bridge in formatting the RELEASING.md. It might be worth doing similar here since it follows a similar process. So for example naming the sections the same and using e.g. `release-<Major>.<Minor>.x` instead of `release-M.N.x`. It will mean that if the process changes in future it should be easier to make similar changes to each one.

Also for those other repos we've put this file under a `development-docs` folder, so again you could do the same here for consistency.

In future maybe we should consider having some of these steps somewhere central but that is a separate issue.

development-docs/RELEASING.md
```diff
@@ -0,0 +1,82 @@
...
+Wait until the build workflow is (successfully) finished for the last commit in the release branch.
+
+Then run the release workflow manually from the GitHub Actions UI:
+1. Go to the **Actions** tab in the GitHub repository
+2. Select the **Release** workflow from the left sidebar
+3. Click **Run workflow** button
+4. Fill in the required parameters:
+   * **Release Version**: for example `1.2.0`
+
+The release workflow will push the images to the registry and publish the Java artifacts to Maven Central.
```
im-konge:
```suggestion
The release workflow will publish the Java artifacts to Maven Central.
```

im-konge (approved):

PR #123, opened 2026-03-24
Add #HELP message metric output

### LINKED ISSUE #105: Counters have a suffix "_total"
The Prometheus library appends the `_total` type suffix to all counter metrics, which you should remove when adding these metrics to the `allowlist`. If you forget or don't know that you have to remove this suffix from the endpoint output, you can easily make a mistake when filtering by the exact metric name (see https://github.com/strimzi/strimzi-kafka-operator/pull/12142).

Prometheus also appends unit suffix (`_bytes`, `_ratio` and `_milliseconds`), so the issue is not limited to the type suffix (`_total`). As a minimum we should document this behavior in the README, but it would be good to export the original metrics name without any change as OTel collector does (https://github.com/open-telemetry/opentelemetry-collector/issues/6403).

OwenCorrigan76:
This PR is designed to fix https://github.com/strimzi/metrics-reporter/issues/105. 
A Help message is added to the metrics to document how a corresponding metric name is to be used in the Metric Reporter configuration in relation to the `allowList`.

### DIFF
--- a/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterTest.java
+++ b/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterTest.java
@@ -126,4 +126,28 @@ public class ClientMetricsReporterTest {
         }
     }
 
+    @Test
+    public void testHelpMessageInOutput() throws Exception {
+        ClientMetricsReporter reporter = new ClientMetricsReporter(registry, kafkaCollector);
+        configs.put(ClientMetricsReporterConfig.ALLOWLIST_CONFIG, "kafka_producer_group_name.*");
+        reporter.configure(configs);
+        reporter.contextChange(new KafkaMetricsContext("kafka.producer"));
+
+        int port = reporter.getPort().orElseThrow();
+
+        // Add a metric that matches the allowlist
+        KafkaMetric metric = newKafkaMetric("name", "group", (config, now) -> 0, LABELS);
+        reporter.metricChange(metric);
+
+        // Get metrics  including comments
+        List<String> metrics = getMetrics(port, true);
+
+        // Verify that the metrics contains the "Use prometheusMetricName in allowlist" help line
+        boolean foundHelpLine = metrics.stream()
+                .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use kafka_producer_group_name") && line.contains("in allowlist"));
+        assertTrue(foundHelpLine, "Expected to find '# HELP' line with 'Use prometheusMetricName in allowlist' message");
+
+        reporter.close();
+    }
+
 }
--- a/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterTest.java
+++ b/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterTest.java
@@ -7,6 +7,7 @@ package io.strimzi.kafka.metrics.prometheus;
 import io.prometheus.metrics.model.registry.PrometheusRegistry;
 import io.strimzi.kafka.metrics.prometheus.common.PrometheusCollector;
 import io.strimzi.kafka.metrics.prometheus.kafka.KafkaCollector;
+import io.strimzi.kafka.metrics.prometheus.kafka.KafkaMetricWrapper;
 import org.apache.kafka.common.metrics.KafkaMetric;
 import org.apache.kafka.common.metrics.KafkaMetricsContext;
 import org.junit.jupiter.api.BeforeEach;
@@ -142,10 +143,12 @@ public class ClientMetricsReporterTest {
         // Get metrics  including comments
         List<String> metrics = getMetrics(port, true);
 
+        String expectedPrometheusName = KafkaMetricWrapper.prometheusName("kafka.producer", metric.metricName());
+
         // Verify that the metrics contains the "Use prometheusMetricName in allowlist" help line
         boolean foundHelpLine = metrics.stream()
-                .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use kafka_producer_group_name") && line.contains("in allowlist"));
-        assertTrue(foundHelpLine, "Expected to find '# HELP' line with 'Use prometheusMetricName in allowlist' message");
+                .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use " + expectedPrometheusName) && line.contains("in allowlist"));
+        assertTrue(foundHelpLine, "Expected to find '# HELP' line with 'Use " + expectedPrometheusName + "in allowlist' message");
 
         reporter.close();
     }
--- a/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterTest.java
+++ b/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterTest.java
@@ -147,7 +147,7 @@ public class ClientMetricsReporterTest {
 
         // Verify that the metrics contains the "Use prometheusMetricName in allowlist" help line
         boolean foundHelpLine = metrics.stream()
-                .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use " + expectedPrometheusName) && line.contains("in allowlist"));
+                .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use " + expectedPrometheusName + " in allowlist"));
         assertTrue(foundHelpLine, "Expected to find '# HELP' line with 'Use " + expectedPrometheusName + "in allowlist' message");
 
         reporter.close();
--- a/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
+++ b/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
@@ -126,32 +126,28 @@ public class KafkaCollectorTest {
         collector.addReporter(reporter);
 
         // Test numeric metric
-        AtomicInteger value = new AtomicInteger(1);
         MetricName numericMetricName = new MetricName("testMetric", "testGroup", "description", tagsMap);
-        MetricWrapper metricWrapper = newKafkaMetricWrapper(numericMetricName, (config, now) -> value.get());
+        MetricWrapper metricWrapper = newKafkaMetricWrapper(numericMetricName, (config, now) -> new AtomicInteger(1).get());
         reporter.addMetric(numericMetricName, metricWrapper);
 
         List<? extends MetricSnapshot> metrics = collector.collect();
         assertEquals(1, metrics.size());
-        MetricSnapshot snapshot = metrics.get(0);
 
         String expectedHelpMessage = "Use " + metricWrapper.prometheusName() + " in allowlist";
-        assertEquals(expectedHelpMessage, snapshot.getMetadata().getHelp());
+        assertEquals(expectedHelpMessage, metrics.get(0).getMetadata().getHelp());
 
         reporter.removeMetric(numericMetricName);
 
         // Test non-numeric metric
-        String nonNumericValue = "testValue";
         MetricName stringMetricName = new MetricName("stringMetric", "testGroup", "description", tagsMap);
-        MetricWrapper stringMetricWrapper = newKafkaMetricWrapper(stringMetricName, (config, now) -> nonNumericValue);
+        MetricWrapper stringMetricWrapper = newKafkaMetricWrapper(stringMetricName, (config, now) -> "testValue");
         reporter.addMetric(stringMetricName, stringMetricWrapper);
 
         metrics = collector.collect();
         assertEquals(1, metrics.size());
-        MetricSnapshot stringSnapshot = metrics.get(0);
 
         expectedHelpMessage = "Use " + stringMetricWrapper.prometheusName() + " in allowlist";
-        assertEquals(expectedHelpMessage, stringSnapshot.getMetadata().getHelp());
+        assertEquals(expectedHelpMessage, metrics.get(0).getMetadata().getHelp());
     }
 
     private MetricWrapper newKafkaMetricWrapper(MetricName metricName, Gauge<?> gauge) {
--- a/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
+++ b/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
@@ -114,6 +114,57 @@ public class KafkaCollectorTest {
         assertEquals(0, metrics.size());
     }
 
+    @Test
+    public void testHelpMessageForMetric() {
+        KafkaCollector collector = new KafkaCollector();
+        AbstractReporter reporter = new AbstractReporter() {
+            @Override
+            protected Pattern allowlist() {
+                return Pattern.compile(".*");
+            }
+        };
+        collector.addReporter(reporter);
+
+        // Adding a numeric metric
+        AtomicInteger value = new AtomicInteger(1);
+        MetricName metricName = new MetricName("testMetric", "testGroup", "description", tagsMap);
+        MetricWrapper metricWrapper = newKafkaMetricWrapper(metricName, (config, now) -> value.get());
+        reporter.addMetric(metricName, metricWrapper);
+
+        List<? extends MetricSnapshot> metrics = collector.collect();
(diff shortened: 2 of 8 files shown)


OwenCorrigan76:
@mimaison I want sure whether the Help message should be the same in `KafkaCollector`?
Also, do we need a unit test to make sure the message is correct?

mimaison:
> Also, do we need a unit test to make sure the message is correct?

Absolutely!

server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerCollector.java
```diff
@@ -93,32 +93,32 @@ public List<MetricSnapshot> collect() {
                 Object metric = metricWrapper.metric();
                 Labels labels = metricWrapper.labels();
                 LOG.debug("Collecting Yammer metric {} with the following labels: {}", prometheusMetricName, labels);
-
+                String helpMessage = String.format("Use '" + prometheusMetricName + "' in allowlist");
```
mimaison: Here you are combining `String.format()` and concatenation with the `+` operator. It makes no sense of using both as you're effectively building the full string with the concatenation but also then calling `String.format()` without any arguments!

You can just use concatenation here.

Same in `YammerCollector`.

mimaison (commented):
Thanks for the PR. We need to adjust the logic building the message. We also want to add tests.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
```diff
@@ -114,6 +114,57 @@ protected Pattern allowlist() {
         assertEquals(0, metrics.size());
     }
 
+    @Test
+    public void testHelpMessageForMetric() {
```
mimaison: I think we could merge these 2 new tests.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
```diff
@@ -114,6 +114,57 @@ protected Pattern allowlist() {
...
+        AtomicInteger value = new AtomicInteger(1);
+        MetricName metricName = new MetricName("testMetric", "testGroup", "description", tagsMap);
+        MetricWrapper metricWrapper = newKafkaMetricWrapper(metricName, (config, now) -> value.get());
+        reporter.addMetric(metricName, metricWrapper);
+
+        List<? extends MetricSnapshot> metrics = collector.collect();
+        assertEquals(1, metrics.size());
+        MetricSnapshot snapshot = metrics.get(0);
+
+        String expectedHelpMessage = "Use kafka_server_testgroup_testmetric in allowlist";
```
mimaison: Instead of fully hard-coding the exact message, can we built it like you did in `YammerCollectorTest`?

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
mimaison: Same in `YammerCollectorTest`

mimaison (commented):
Thanks for the PR. I left some suggestions

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
OwenCorrigan76: Done

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
OwenCorrigan76: Yes. Had meant to do that in both but missed it. Thanks

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
```diff
@@ -114,6 +114,46 @@ protected Pattern allowlist() {
...
+        AbstractReporter reporter = new AbstractReporter() {
+            @Override
+            protected Pattern allowlist() {
+                return Pattern.compile(".*");
+            }
+        };
+        collector.addReporter(reporter);
+
+        // Test numeric metric
+        AtomicInteger value = new AtomicInteger(1);
```
mimaison: Since we don't use this value later in the test, we can inline this variable.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
```diff
@@ -114,6 +114,46 @@ protected Pattern allowlist() {
...
+        assertEquals(1, metrics.size());
+        MetricSnapshot snapshot = metrics.get(0);
+
+        String expectedHelpMessage = "Use " + metricWrapper.prometheusName() + " in allowlist";
+        assertEquals(expectedHelpMessage, snapshot.getMetadata().getHelp());
+
+        reporter.removeMetric(numericMetricName);
+
+        // Test non-numeric metric
+        String nonNumericValue = "testValue";
```
mimaison: Same here, we can inline this variable.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
```diff
@@ -114,6 +114,46 @@ protected Pattern allowlist() {
...
+        MetricName stringMetricName = new MetricName("stringMetric", "testGroup", "description", tagsMap);
+        MetricWrapper stringMetricWrapper = newKafkaMetricWrapper(stringMetricName, (config, now) -> nonNumericValue);
+        reporter.addMetric(stringMetricName, stringMetricWrapper);
+
+        metrics = collector.collect();
+        assertEquals(1, metrics.size());
+        MetricSnapshot stringSnapshot = metrics.get(0);
```
mimaison: This variable can be inlined too

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
```diff
@@ -114,6 +114,46 @@ protected Pattern allowlist() {
...
+
+        // Test numeric metric
+        AtomicInteger value = new AtomicInteger(1);
+        MetricName numericMetricName = new MetricName("testMetric", "testGroup", "description", tagsMap);
+        MetricWrapper metricWrapper = newKafkaMetricWrapper(numericMetricName, (config, now) -> value.get());
+        reporter.addMetric(numericMetricName, metricWrapper);
+
+        List<? extends MetricSnapshot> metrics = collector.collect();
+        assertEquals(1, metrics.size());
+        MetricSnapshot snapshot = metrics.get(0);
```
mimaison: We can also inline this variable

server-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerCollectorTest.java
```diff
@@ -102,6 +102,45 @@ protected Pattern allowlist() {
         assertInfoSnapshot(snapshot, labels, "name", nonNumericValue);
     }
 
+    @Test
```
mimaison, line 105: The comments in `KafkaCollectorTest` also apply to this file.

mimaison (commented):
Thanks for the update! I made another pass and left a few more suggestions.

OwenCorrigan76:
Thanks a mil @mimaison. Will update shortly.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
```diff
@@ -114,6 +114,42 @@ protected Pattern allowlist() {
...
+            @Override
+            protected Pattern allowlist() {
+                return Pattern.compile(".*");
+            }
+        };
+        collector.addReporter(reporter);
+
+        // Test numeric metric
+        MetricName numericMetricName = new MetricName("testMetric", "testGroup", "description", tagsMap);
+        MetricWrapper metricWrapper = newKafkaMetricWrapper(numericMetricName, (config, now) -> new AtomicInteger(1).get());
```
mimaison: No need for an `AtomicInteger` here, you can just do `(config, now) -> 1`

mimaison: Same in YammerCollectorTest

mimaison (commented):
Thanks for the updates. Just one more small change and we should be good.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/kafka/KafkaCollectorTest.java
OwenCorrigan76: @mimaison Thanks. Updated now.

mimaison (approved):
LGTM

scholzj (commented):
* Do we have any integration tests where we can test the comments in the actual metrics output from the HTTP server?
* If we already do this, should we add more information to it? Such as the underlying type? Some metric description

@OwenCorrigan76 Can you share some example of how the metric output looks like?

OwenCorrigan76:
@scholzj 
This is an example of how the metrics output looks:

```
# HELP kafka_controller_controllerstats_electionfromeligibleleaderreplicaspersec_total Use 'kafka_controller_controllerstats_electionfromeligibleleaderreplicaspersec' in allowlist
# TYPE kafka_controller_controllerstats_electionfromeligibleleaderreplicaspersec_total counter
kafka_controller_controllerstats_electionfromeligibleleaderreplicaspersec_total 0.0
```

As for the IT's, looking....

scholzj:
@OwenCorrigan76 Ok, so the TYPE is already there, great. I did not realized that. I think the integration test where it actually checks the output would be nice. But if there isn't one already, I can probably live without it.

OwenCorrigan76:
@scholzj @mimaison 
Does this suffice as part of the integration tests which use MetricUtils?
https://github.com/strimzi/metrics-reporter/blob/main/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java#L76-L78

mimaison:
No. This does not assert anything about the help message. It does not even assert there are lines starting with `#`

OwenCorrigan76:
Is the test that should have an assertion about the help message?
And equivelant in consumer and streams??
https://github.com/strimzi/metrics-reporter/blob/main/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/integration/TestProducerMetricsIT.java#L71

mimaison:
> Is the test that should have an assertion about the help message?
> And equivelant in consumer and streams??
> https://github.com/strimzi/metrics-reporter/blob/main/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/integration/TestProducerMetricsIT.java#L71

I don't understand what you mean.

Maybe something in `ClientMetricsReporterTest`, `ServerKafkaMetricsReporterTest` and `ServerYammerMetricsReporterTest` that checks we get the "Use ..." lines in the output would be fine.

scholzj:
> Is the test that should have an assertion about the help message? And equivelant in consumer and streams?? https://github.com/strimzi/metrics-reporter/blob/main/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/integration/TestProducerMetricsIT.java#L71

So, I do not think you can have it in this exact test. But it should be similar. Just in the _assert_, you should check that the output does have some specific HELM comments. The classes mentioned by Mickael seem to also do something similar. So if you can add it to Unit tests nstead of Integration tests, it would be probably better.

(However, the `MetricsUtils.getMetrics` actually filters out the comments. So I do not think you can use that easily. You might need a separate method that will nto filter them)

mimaison:
You could overload `MetricsUtils.getMetrics()` with an extra boolean argument whether to filter out comments.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterTest.java
```diff
@@ -126,4 +126,28 @@ public void testContextChange() {
...
+        // Add a metric that matches the allowlist
+        KafkaMetric metric = newKafkaMetric("name", "group", (config, now) -> 0, LABELS);
+        reporter.metricChange(metric);
+
+        // Get metrics  including comments
+        List<String> metrics = getMetrics(port, true);
+
+        // Verify that the metrics contains the "Use prometheusMetricName in allowlist" help line
+        boolean foundHelpLine = metrics.stream()
+                .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use kafka_producer_group_name") && line.contains("in allowlist"));
```
mimaison: Instead of hard coding `kafka_producer_group_name` we can compute it with `KafkaMetricWrapper.prometheusName("kafka.producer", metric.metricName());`

mimaison: We should be able to do something similar in the server tests

mimaison (commented):
The tests look fine. I left a small suggestion

scholzj (approved):
LGTM. Thanks.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterTest.java
```diff
@@ -126,4 +127,30 @@ public void testContextChange() {
...
+        reporter.metricChange(metric);
+
+        // Get metrics  including comments
+        List<String> metrics = getMetrics(port, true);
+
+        String expectedPrometheusName = KafkaMetricWrapper.prometheusName("kafka.producer", metric.metricName());
+
+        // Verify that the metrics contains the "Use prometheusMetricName in allowlist" help line
+        boolean foundHelpLine = metrics.stream()
+                .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use " + expectedPrometheusName) && line.contains("in allowlist"));
```
mimaison: We can simplify this into:
```
boolean foundHelpLine = metrics.stream()
                .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use " + expectedPrometheusName + " in allowlist"));
```

mimaison: Same in the server tests

OwenCorrigan76: Of course. Removed unnecessary concatenation

mimaison: You are missing a space before `in allowlist`

server-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ServerYammerMetricsReporterTest.java
```diff
@@ -108,4 +110,34 @@ private void removeMetric(String group, String type, String name) {
...
+            newCounter("group", "type", "name");
+
+            // Get metrics output including comments
+            List<String> metrics = getMetrics(port, true);
+
+            String expectedPrometheusName = YammerMetricWrapper.prometheusName(metricName);
+
+            // Verify that the output contains the "Use prometheusMetricName in allowlist" help line
+            boolean foundHelpLine = metrics.stream()
+                    .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use " + YammerMetricWrapper.prometheusName(metricName)) && line.contains("in allowlist"));
```
mimaison: You can do the simplification here too:
```
boolean foundHelpLine = metrics.stream()
                    .anyMatch(line -> line.startsWith("# HELP") && line.contains("Use " + YammerMetricWrapper.prometheusName(metricName) + " in allowlist"));
```

OwenCorrigan76: Done. Sorry for the extra work reviewing. I should have spotted these.

PR #125, opened 2026-04-17
Replace developer list with Strimzi Project Maintainers

mimaison:
Update the developers block to align with other Strimzi projects

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,70 +29,10 @@
 
   <developers>
     <developer>
-      <name>Tom Bentley</name>
-      <email>tbentley@redhat.com</email>
-      <organization>Red Hat</organization>
-      <organizationUrl>https://www.redhat.com</organizationUrl>
-    </developer>
-    <developer>
-      <name>Paolo Patierno</name>
-      <email>ppatierno@live.com</email>
-      <organization>Red Hat</organization>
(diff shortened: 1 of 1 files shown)


scholzj (approved):
I'm fine if you want to merge it. But it is actually not correct here as it takes away your credit @mimaison. If you want feel free to keep your self in the list explicitely. Or we can change it to something like "Strimzi Project Maintainers and Component owners", or something liek that. WDYT @ppatierno?

ppatierno:
@scholzj I agree with you, @mimaison should be in the developers list. So I guess we have two options here: adding him explicitly alongside the change just applied in the PR or as you said we could use more "Strimzi Project Maintainers and Component owners" but in this case we should do the same across other repos if we want to be consistent and the email won't reflect exactly the naming (because the mailing list is only for maintainers). I don't have a really strong opinion both are fine with me but maybe doing as you suggest would be less work in the future with the addition of more component owners.

scholzj:
@ppatierno Yeah ... `Strimzi Project Maintainers and Component owners` is a bit of a hack 🤷. I guess for me it is mostly up to @mimaison if he just wants to add his name next to the maintainers or not 🤷.

mimaison:
I don't mind not being explicitly listed, I already receive enough emails

ppatierno (approved):

PR #128, opened 2026-05-19
Fix label values by removing prometheus sanitation for label values

### LINKED ISSUE #127: Label sanitization leads to altered topic names in metric labels
It seems like Strimzi metrics reporter applies PrometheusNaming.sanitizeMetricName not only to metric names but also to label values. This can be misleading, especially for labels that contain user-defined identifiers such as Kafka topic names.

For example, a topic name like:

`env.topicname.version` is transformed into:`kafka_server_brokertopicmetrics_messagesinpersec_total{opic="env_topicname_version"}`

Users may rely on exact naming (e.g., for topic-based filtering, dashboards, or alerts). From our perspective, metric label values should preserve their original form.

Only metric names should be sanitized to avoid modifying label values. Alternatively, providing a configuration option to disable label sanitization for labels would also be a viable solution.

leisma:
Closes: https://github.com/strimzi/metrics-reporter/issues/127

Tested the metrics-reporter in our lab environment and verified it works as expected.

### DIFF
--- a/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
+++ b/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
@@ -75,11 +75,6 @@ public class Listener {
      */
     public static class ListenerValidator implements ConfigDef.Validator {
 
-        /**
-         * Creates a new ListenerValidator
-         */
-        public ListenerValidator() {}
-
         @Override
         public void ensureValid(String name, Object value) {
             parseListener(String.valueOf(value));
--- a/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
+++ b/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
@@ -75,6 +75,11 @@ public class Listener {
      */
     public static class ListenerValidator implements ConfigDef.Validator {
 
+        /**
+         * Creates a new ListenerValidator
+         */
+        public ListenerValidator() {}
+
         @Override
         public void ensureValid(String name, Object value) {
             parseListener(String.valueOf(value));
--- a/server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
+++ b/server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
@@ -57,7 +57,9 @@ public class YammerMetricWrapper extends MetricWrapper {
                 labelKeys.add(parts[i]);
             }
         }
-        if (labelKeys.isEmpty() || mbeanName == null) return Labels.EMPTY;
+        if (labelKeys.isEmpty() || mbeanName == null) {
+            return Labels.EMPTY;
+        }
         Labels.Builder builder = Labels.builder();
         Set<String> labelNames = new HashSet<>();
         // Example mbeanName: "kafka.server:type=BrokerTopicMetrics,name=MessagesInPerSec,topic=env.topicname.version"
--- a/server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
+++ b/server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
@@ -26,12 +26,12 @@ public class YammerMetricWrapper extends MetricWrapper {
     /**
      * Constructor from Yammer Metrics
      * @param prometheusName The name of the metric in the prometheus format
-     * @param scope The scope of the Yammer metric
+     * @param metricName The Yammer MetricName
      * @param metric The Yammer metric
      * @param attribute The attribute of the Yammer metric
      */
-    public YammerMetricWrapper(String prometheusName, String scope, Metric metric, String attribute) {
-        super(prometheusName, labelsFromScope(scope, prometheusName), metric, attribute);
+    public YammerMetricWrapper(String prometheusName, MetricName metricName, Metric metric, String attribute) {
+        super(prometheusName, labelsFromScopeAndMBeanName(metricName.getScope(), metricName.getMBeanName(), prometheusName), metric, attribute);
     }
 
     /**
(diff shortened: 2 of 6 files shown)


mimaison:
Thanks for the PR. I'm a bit busy this week but hopefully I'll be able to take a look next week.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
```diff
@@ -75,6 +75,11 @@ public int hashCode() {
      */
     public static class ListenerValidator implements ConfigDef.Validator {
 
+        /**
+         * Creates a new ListenerValidator
+         */
+        public ListenerValidator() {}
+
```
mimaison: How is this related to the issue?

server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
```diff
@@ -47,19 +47,34 @@ public static String prometheusName(MetricName metricName) {
                         metricName.getName()).toLowerCase(Locale.ROOT));
     }
 
-    static Labels labelsFromScope(String scope, String metricName) {
-        Labels.Builder builder = Labels.builder();
-        Set<String> labelNames = new HashSet<>();
+    static Labels labelsFromScopeAndMBeanName(String scope, String mbeanName, String metricName) {
+        // Example scope: "type.kafka.name.BytesInPerSec"
```
mimaison: The comments are consistent but don't expose a valid example.
An example scope would be `topic.env_topicname_version`, then on line 71, the comment would say that `type` in not in scope thus ignored.

server-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapperTest.java
```diff
@@ -18,17 +17,23 @@
...
-        assertEquals(Labels.EMPTY, YammerMetricWrapper.labelsFromScope(null, "name"));
-        assertEquals(Labels.EMPTY, YammerMetricWrapper.labelsFromScope("k1", "name"));
-        assertEquals(Labels.EMPTY, YammerMetricWrapper.labelsFromScope("k1.", "name"));
-        assertEquals(Labels.EMPTY, YammerMetricWrapper.labelsFromScope("k1.v1.k", "name"));
-
-        Labels labels = YammerMetricWrapper.labelsFromScope("k-1.v1.k_1.v2", "name");
-        assertEquals("k_1", PrometheusNaming.sanitizeLabelName("k-1"));
-        assertEquals("v1", labels.get("k_1"));
-        assertEquals(1, labels.size());
+    public void testLabelsFromScopeAndMBeanName() {
```
mimaison, line 21: We seem to have lost the test case checking for duplicate labels

mimaison (commented):
Thanks for the PR! The logic seems fine, I left a few comments.
Also We should probably check that behavior in TestServerMetricsIT, it should be easy to create a topic with dots in its name in one of the tests.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
leisma: Not really related to this issue, but it fixes a lint warning from xdoclint. Should I remove it?

server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
leisma: I fixed the comments and should now provide a valid example.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
mimaison: Yes, if we want to address it, it's better to do it in a separate PR.

server-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapperTest.java
leisma: Yes you are right! I re-added the tests.

leisma:
Hey @mimaison, thanks for the review! I tried to address all your feedback and also added an integration test that verifies that topic names with dots are preserved in the metrics.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
leisma: I removed the fix.

server-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/integration/TestServerMetricsIT.java
```diff
@@ -154,6 +159,28 @@ public void testReconfigureAllowlist() throws Exception {
...
+        }
+
+        try (KafkaProducer<String, String> producer = new KafkaProducer<>(Map.of(
+                ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, cluster.getBootstrapServers(),
+                ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName(),
+                ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName()))) {
+            producer.send(new ProducerRecord<>(topic, "key", "value")).get();
+        }
+
+        List<String> patterns = List.of("kafka_server_brokertopicmetrics_.*topic=\"env\\.topicname\\.version\".*");
```
mimaison: There's no need to escape the dots. We could even do:
```
List<String> patterns = List.of("kafka_server_brokertopicmetrics_.*topic=\"" + topic + "\".*");
```

mimaison:
Thanks for the updates!

It looks like you did not signed off your last commit, hence why the DCO check is failing. 
You can run `git commit -s --amend`  to sign your last commit and then `git push -f` to update your branch.

mimaison (approved):
Thanks for the PR, LGTM

server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
```diff
@@ -47,19 +47,34 @@ public static String prometheusName(MetricName metricName) {
...
+        // Example mbeanName: "kafka.server:type=BrokerTopicMetrics,name=MessagesInPerSec,topic=env.topicname.version"
+        int colonIdx = mbeanName.indexOf(':');
+        if (colonIdx >= 0) {
+            for (String property : mbeanName.substring(colonIdx + 1).split(",")) {
+                int eqIdx = property.indexOf('=');
+                if (eqIdx < 0) continue;
+                String key = property.substring(0, eqIdx);
+                String value = property.substring(eqIdx + 1);
+                // Example labels: {topic="env.topicname.version"} - type and name properties are ignored as not in scope
+                if (!labelKeys.contains(key)) continue;
```
ppatierno, line 74: I was wondering if it's not better to go the opposite logic in the `if` condition and then avoiding the `continue` which I find closer to "spaghetti" code, unless really necessary. But it's just my opinion.

server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
```diff
@@ -47,19 +47,34 @@ public static String prometheusName(MetricName metricName) {
...
-                        builder.label(newLabelName, parts[i + 1]);
-                    } else {
-                        LOG.warn("Ignoring duplicate label key: {} with value: {} from metric: {} ", newLabelName, parts[i + 1], metricName);
-                    }
+            for (int i = 0; i < parts.length - 1; i += 2) {
+                // Example labelKeys = {"topic"}
+                labelKeys.add(parts[i]);
+            }
+        }
+        if (labelKeys.isEmpty() || mbeanName == null) return Labels.EMPTY;
```
ppatierno: I think a `return` statement should be pretty visible and having it on a single line also behind an `if` drives people to lose it. I would put with right indentation on the next line imho.

server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
leisma: Thanks for the feedback! I would keep it as it is, the continue here is an intentional guard clause, consistent with line 68, and it keeps the happy path flat without adding another indentation level. But I can change if you don't have a good feeling with it!

server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
leisma: Good point, you're right! I just fixed it.

server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
ppatierno: Not that much a fan but if Mickael was fine, no need to change.

ppatierno (approved):
LGTM.

server-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/yammer/YammerMetricWrapper.java
mimaison: Yes it's not ideal but the alternative isn't much better. Since we already have a `continue` above, I'm ok to keep the code as-is, at least for now.

PR #130, opened 2026-05-26
feat: Add support for HTTPS

### LINKED ISSUE #65: Add support for https
The metrics reporter should support exposing the metrics endpoint via https.

antonio-pedro99:
This PR fixes https://github.com/strimzi/metrics-reporter/issues/65 implementing https://github.com/strimzi/proposals/pull/201.

Disclaimer: Codex was used for testing and boilerplates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,25 @@ After building, make sure the metrics reporter JARs located under `target/metric
 
 The metrics reporter has the following configurations:
 
-- `prometheus.metrics.reporter.listener`: The HTTP listener to expose the metrics. It must be in the `http://[host]:[port]` format. This defaults to `http://:8080`.
+- `prometheus.metrics.reporter.listener`: The HTTP or HTTPS listener to expose the metrics. It must be in the `http://[host]:[port]` or `https://[host]:[port]` format. This defaults to `http://:8080`.
 - `prometheus.metrics.reporter.listener.enable`: Enable the listener to expose the metrics. This defaults to `true`.
 - `prometheus.metrics.reporter.allowlist`: A comma separated list of regex patterns to specify the metrics to collect. This defaults to `.*`.
+- `prometheus.metrics.reporter.listener.ssl.certificate.location`: The path to the PEM file containing the server certificate or certificate chain.
+- `prometheus.metrics.reporter.listener.ssl.key.location`: The path to the PEM file containing the unencrypted private key.
+- `prometheus.metrics.reporter.listener.ssl.certificate`: The server certificate or certificate chain in PEM format, provided as an inline string. This takes precedence over `prometheus.metrics.reporter.listener.ssl.certificate.location`.
+- `prometheus.metrics.reporter.listener.ssl.key`: The unencrypted private key in PEM format, provided as an inline string. This takes precedence over `prometheus.metrics.reporter.listener.ssl.key.location`.
+- `prometheus.metrics.reporter.listener.ssl.enabled.protocols`: A comma separated list of enabled secure transport protocols. This defaults to `TLSv1.2,TLSv1.3`.
+- `prometheus.metrics.reporter.listener.ssl.enabled.cipher.suites`: A comma separated list of cipher suites that the server will support. Empty uses Java's default cipher suites.
+
+To expose metrics over HTTPS, configure the listener with the `https://` scheme and provide the server certificate and private key:
+
+```properties
+prometheus.metrics.reporter.listener=https://:8443
+prometheus.metrics.reporter.listener.ssl.certificate.location=/path/to/tls.crt
+prometheus.metrics.reporter.listener.ssl.key.location=/path/to/tls.key
+```
+
+The HTTPS listener accepts PEM certificate chains and unencrypted PEM private keys. If the TLS configuration is invalid, such as a missing certificate, unsupported key format, mismatched key, unsupported protocol, or unsupported cipher suite, the reporter fails to start with a configuration error.
 
 ## Running
 
--- a/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
+++ b/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
@@ -58,8 +58,13 @@ public class ClientMetricsReporter extends AbstractReporter implements MetricsRe
         kafkaCollector.addReporter(this);
     }
 
-    // for testing
-    ClientMetricsReporter(PrometheusRegistry registry, KafkaCollector kafkaCollector) {
+    /**
+     * Constructor
+     *
+     * @param registry The Prometheus registry.
+     * @param kafkaCollector The Kafka metrics collector.
+     */
+    public ClientMetricsReporter(PrometheusRegistry registry, KafkaCollector kafkaCollector) {
         this.registry = registry;
         this.kafkaCollector = kafkaCollector;
         kafkaCollector.addReporter(this);
@@ -120,8 +125,12 @@ public class ClientMetricsReporter extends AbstractReporter implements MetricsRe
         this.prefix = PrometheusNaming.prometheusName(prefix);
     }
 
-    // for testing
-    Optional<Integer> getPort() {
+    /**
+     * Gets the port used by the HTTP server.
+     *
+     * @return The HTTP server port, or empty if the HTTP server is not running.
+     */
+    public Optional<Integer> getPort() {
         return httpServer.map(HttpServers.ServerCounter::port);
     }
 
--- a/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
+++ b/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
@@ -61,10 +61,10 @@ public class ClientMetricsReporter extends AbstractReporter implements MetricsRe
     /**
      * Constructor
      *
-     * @param registry The Prometheus registry.
+     * @param registry       The Prometheus registry.
      * @param kafkaCollector The Kafka metrics collector.
      */
-    public ClientMetricsReporter(PrometheusRegistry registry, KafkaCollector kafkaCollector) {
+    ClientMetricsReporter(PrometheusRegistry registry, KafkaCollector kafkaCollector) {
         this.registry = registry;
         this.kafkaCollector = kafkaCollector;
         kafkaCollector.addReporter(this);
@@ -130,7 +130,7 @@ public class ClientMetricsReporter extends AbstractReporter implements MetricsRe
      *
      * @return The HTTP server port, or empty if the HTTP server is not running.
      */
-    public Optional<Integer> getPort() {
+    Optional<Integer> getPort() {
         return httpServer.map(HttpServers.ServerCounter::port);
     }
 
--- a/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java
+++ b/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java
@@ -4,10 +4,13 @@
  */
 package io.strimzi.kafka.metrics.prometheus;
 
+import com.sun.net.httpserver.HttpsConfigurator;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.prometheus.metrics.model.registry.PrometheusRegistry;
 import io.strimzi.kafka.metrics.prometheus.http.HttpServers;
+import io.strimzi.kafka.metrics.prometheus.http.HttpsConfiguratorFactory;
 import io.strimzi.kafka.metrics.prometheus.http.Listener;
+import io.strimzi.kafka.metrics.prometheus.http.SSLContextFactory;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
@@ -38,7 +41,65 @@ public class ClientMetricsReporterConfig extends AbstractConfig {
      * Default value for the listener configuration.
      */
     public static final String LISTENER_CONFIG_DEFAULT = "http://:8080";
-    private static final String LISTENER_CONFIG_DOC = "The HTTP listener to expose the metrics.";
+    private static final String LISTENER_CONFIG_DOC = "The HTTP or HTTPS listener to expose the metrics.";
+    private static final String LISTENER_SSL_CONFIG_PREFIX = LISTENER_CONFIG + ".ssl.";
+
+    /**
+     * Configuration key for the PEM file containing the server certificate or certificate chain.
+     */
+    public static final String LISTENER_SSL_CERTIFICATE_LOCATION_CONFIG =
+            LISTENER_SSL_CONFIG_PREFIX + "certificate.location";
+    private static final String LISTENER_SSL_CERTIFICATE_LOCATION_CONFIG_DOC =
+            "The path to the PEM file containing the server certificate or certificate chain.";
+
+    /**
+     * Configuration key for the PEM file containing the server private key.
+     */
+    public static final String LISTENER_SSL_KEY_LOCATION_CONFIG = LISTENER_SSL_CONFIG_PREFIX + "key.location";
+    private static final String LISTENER_SSL_KEY_LOCATION_CONFIG_DOC =
+            "The path to the PEM file containing the server private key.";
+
+    /**
+     * Configuration key for the inline PEM server certificate or certificate chain.
+     */
+    public static final String LISTENER_SSL_CERTIFICATE_CONFIG = LISTENER_SSL_CONFIG_PREFIX + "certificate";
+    private static final String LISTENER_SSL_CERTIFICATE_CONFIG_DOC =
+            "The server certificate or certificate chain in PEM format. This takes precedence over " +
+                    LISTENER_SSL_CERTIFICATE_LOCATION_CONFIG + ".";
+
+    /**
+     * Configuration key for the inline PEM server private key.
+     */
+    public static final String LISTENER_SSL_KEY_CONFIG = LISTENER_SSL_CONFIG_PREFIX + "key";
+    private static final String LISTENER_SSL_KEY_CONFIG_DOC =
+            "The server private key in PEM format. This takes precedence over " +
+                    LISTENER_SSL_KEY_LOCATION_CONFIG + ".";
+
+    /**
+     * Configuration key for the enabled secure transport protocols.
+     */
+    public static final String LISTENER_SSL_ENABLED_PROTOCOLS_CONFIG = LISTENER_SSL_CONFIG_PREFIX + "enabled.protocols";
+
+    /**
+     * Default value for the enabled secure transport protocols.
+     */
+    public static final String LISTENER_SSL_ENABLED_PROTOCOLS_CONFIG_DEFAULT = "TLSv1.2,TLSv1.3";
+    private static final String LISTENER_SSL_ENABLED_PROTOCOLS_CONFIG_DOC =
+            "A comma separated list of enabled secure transport protocols.";
+
+    /**
+     * Configuration key for the enabled cipher suites.
+     */
+    public static final String LISTENER_SSL_ENABLED_CIPHER_SUITES_CONFIG =
+            LISTENER_SSL_CONFIG_PREFIX + "enabled.cipher.suites";
+
+    /**
+     * Default value for the enabled cipher suites. An empty list uses Java's default cipher suites.
+     */
+    public static final String LISTENER_SSL_ENABLED_CIPHER_SUITES_CONFIG_DEFAULT = "";
+    private static final String LISTENER_SSL_ENABLED_CIPHER_SUITES_CONFIG_DOC =
+            "A comma separated list of cipher suites that the server will support. " +
+                    "Empty uses Java's default cipher suites.";
 
     /**
      * Configuration key to determine if the listener is enabled or not.
@@ -60,17 +121,37 @@ public class ClientMetricsReporterConfig extends AbstractConfig {
      * Default value for the allowlist configuration.
      */
     public static final String ALLOWLIST_CONFIG_DEFAULT = ".*";
-    private static final String ALLOWLIST_CONFIG_DOC = "A comma separated list of regex patterns to specify the metrics to collect.";
+    private static final String ALLOWLIST_CONFIG_DOC =
+            "A comma separated list of regex patterns to specify the metrics to collect.";
 
     static final ConfigDef CONFIG_DEF = new ConfigDef()
-            .define(LISTENER_CONFIG, ConfigDef.Type.STRING, LISTENER_CONFIG_DEFAULT, new Listener.ListenerValidator(), ConfigDef.Importance.HIGH, LISTENER_CONFIG_DOC)
-            .define(ALLOWLIST_CONFIG, ConfigDef.Type.LIST, ALLOWLIST_CONFIG_DEFAULT, ConfigDef.Importance.HIGH, ALLOWLIST_CONFIG_DOC)
-            .define(LISTENER_ENABLE_CONFIG, ConfigDef.Type.BOOLEAN, LISTENER_ENABLE_CONFIG_DEFAULT, ConfigDef.Importance.HIGH, LISTENER_ENABLE_CONFIG_DOC);
+            .define(LISTENER_CONFIG, ConfigDef.Type.STRING, LISTENER_CONFIG_DEFAULT, new Listener.ListenerValidator(),
+                    ConfigDef.Importance.HIGH, LISTENER_CONFIG_DOC)
+            .define(LISTENER_SSL_CERTIFICATE_LOCATION_CONFIG, ConfigDef.Type.STRING, null,
+                    ConfigDef.Importance.MEDIUM, LISTENER_SSL_CERTIFICATE_LOCATION_CONFIG_DOC)
+            .define(LISTENER_SSL_KEY_LOCATION_CONFIG, ConfigDef.Type.STRING, null,
+                    ConfigDef.Importance.MEDIUM, LISTENER_SSL_KEY_LOCATION_CONFIG_DOC)
+            .define(LISTENER_SSL_CERTIFICATE_CONFIG, ConfigDef.Type.PASSWORD, null,
+                    ConfigDef.Importance.MEDIUM, LISTENER_SSL_CERTIFICATE_CONFIG_DOC)
+            .define(LISTENER_SSL_KEY_CONFIG, ConfigDef.Type.PASSWORD, null,
+                    ConfigDef.Importance.MEDIUM, LISTENER_SSL_KEY_CONFIG_DOC)
+            .define(LISTENER_SSL_ENABLED_PROTOCOLS_CONFIG, ConfigDef.Type.LIST,
+                    LISTENER_SSL_ENABLED_PROTOCOLS_CONFIG_DEFAULT,
+                    ConfigDef.Importance.MEDIUM, LISTENER_SSL_ENABLED_PROTOCOLS_CONFIG_DOC)
+            .define(LISTENER_SSL_ENABLED_CIPHER_SUITES_CONFIG, ConfigDef.Type.LIST,
+                    LISTENER_SSL_ENABLED_CIPHER_SUITES_CONFIG_DEFAULT,
+                    ConfigDef.Importance.MEDIUM, LISTENER_SSL_ENABLED_CIPHER_SUITES_CONFIG_DOC)
+            .define(ALLOWLIST_CONFIG, ConfigDef.Type.LIST, ALLOWLIST_CONFIG_DEFAULT,
+                    ConfigDef.Importance.HIGH, ALLOWLIST_CONFIG_DOC)
+            .define(LISTENER_ENABLE_CONFIG, ConfigDef.Type.BOOLEAN, LISTENER_ENABLE_CONFIG_DEFAULT,
+                    ConfigDef.Importance.HIGH, LISTENER_ENABLE_CONFIG_DOC);
 
     final Listener listener;
     final boolean listenerEnabled;
     final PrometheusRegistry registry;
     final Pattern allowlist;
+    SSLContextFactory sslContextFactory = null;
+    HttpsConfiguratorFactory httpsConfiguratorFactory = null;
 
     /**
      * Constructor.
@@ -85,6 +166,20 @@ public class ClientMetricsReporterConfig extends AbstractConfig {
         this.allowlist = compileAllowlist(getList(ALLOWLIST_CONFIG));
         this.listenerEnabled = getBoolean(LISTENER_ENABLE_CONFIG);
         this.registry = registry;
+        if (listener.isHttps()) {
+            this.sslContextFactory = new SSLContextFactory(
+                    getString(LISTENER_SSL_CERTIFICATE_LOCATION_CONFIG),
+                    getString(LISTENER_SSL_KEY_LOCATION_CONFIG),
+                    passwordValue(LISTENER_SSL_CERTIFICATE_CONFIG),
+                    passwordValue(LISTENER_SSL_KEY_CONFIG));
+            this.httpsConfiguratorFactory = new HttpsConfiguratorFactory(
+                    getList(LISTENER_SSL_ENABLED_PROTOCOLS_CONFIG),
+                    getList(LISTENER_SSL_ENABLED_CIPHER_SUITES_CONFIG));
+        }
+    }
+
+    private String passwordValue(String config) {
+        return getPassword(config) != null ? getPassword(config).value() : null;
     }
 
     /**
@@ -140,6 +235,8 @@ public class ClientMetricsReporterConfig extends AbstractConfig {
         return "ClientMetricsReporterConfig{" +
                 "listener=" + listener +
                 ", listenerEnabled=" + listenerEnabled +
+                ", sslContextFactory=" + sslContextFactory +
+                ", httpsConfiguratorFactory=" + httpsConfiguratorFactory +
                 ", allowlist=" + allowlist +
                 '}';
     }
@@ -154,8 +251,12 @@ public class ClientMetricsReporterConfig extends AbstractConfig {
             LOG.info("HTTP server listener not enabled");
             return Optional.empty();
         }
-        HttpServers.ServerCounter server = HttpServers.getOrCreate(listener, registry);
-        LOG.info("HTTP server listening on http://{}:{}", listener.host, server.port());
+        HttpsConfigurator httpsConfigurator = null;
+        if (listener.isHttps()) {
+            httpsConfigurator = httpsConfiguratorFactory.create(sslContextFactory.create());
+        }
+        HttpServers.ServerCounter server = HttpServers.getOrCreate(listener, registry, httpsConfigurator);
+        LOG.info("HTTP server listening on {}://{}:{}", listener.scheme, listener.host, server.port());
         return Optional.of(server);
     }
 }
--- a/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java
+++ b/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java
@@ -150,8 +150,8 @@ public class ClientMetricsReporterConfig extends AbstractConfig {
     final boolean listenerEnabled;
     final PrometheusRegistry registry;
     final Pattern allowlist;
-    SSLContextFactory sslContextFactory = null;
-    HttpsConfiguratorFactory httpsConfiguratorFactory = null;
+    final SSLContextFactory sslContextFactory;
+    final HttpsConfiguratorFactory httpsConfiguratorFactory;
 
     /**
      * Constructor.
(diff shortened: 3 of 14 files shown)


client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/HttpServers.java
```diff
@@ -33,9 +35,23 @@ private HttpServers() {
      * @return A ServerCounter instance
      */
     public synchronized static ServerCounter getOrCreate(Listener listener, PrometheusRegistry registry) {
+        return getOrCreate(listener, registry, null);
```
mimaison, line 38: It looks like this method is now unused

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/HttpServers.java
```diff
@@ -4,6 +4,8 @@
  */
 package io.strimzi.kafka.metrics.prometheus.http;
 
+import com.sun.net.httpserver.HttpsConfigurator;
+import edu.umd.cs.findbugs.annotations.Nullable;
```
mimaison, line 8: We don't use this annotation anywhere else in the code base. I wonder if we really need it, WDYT?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
```diff
@@ -30,54 +30,83 @@ public class Listener {
...
 
     /**
      * Validator to check the user provided listener configuration
      */
     public static class ListenerValidator implements ConfigDef.Validator {
 
+        /**
+         * Constructor.
+         */
+        public ListenerValidator() { }
```
mimaison, line 102: Why are we adding this new constructor?

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfigTest.java
```diff
@@ -42,6 +55,81 @@ public void testOverrides() {
...
         assertTrue(config.isAllowed("kafka_server_metric"));
+
+        props = Map.of(LISTENER_CONFIG, "https://:0");
+        config = new ClientMetricsReporterConfig(props, new PrometheusRegistry());
+
+        assertEquals("https://:0", config.listener());
+    }
+
+    @Test
+    public void testHttpListenerDoesNotCreateSslFactories() {
```
mimaison, line 66: Why are we using plural `Factories` instead of `Factory`?

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfigTest.java
```diff
@@ -42,6 +55,81 @@ public void testOverrides() {
...
+
+        assertNull(config.sslContextFactory);
+        assertNull(config.httpsConfiguratorFactory);
+    }
+
+    @Test
+    public void testHttpsListenerCreatesSslFactoriesWithDefaults() {
+        ClientMetricsReporterConfig config = new ClientMetricsReporterConfig(
+                Map.of(LISTENER_CONFIG, "https://:0"),
+                new PrometheusRegistry());
```
mimaison, line 77: Is this a valid SSL configuration? If not, should the constructor throw?

mimaison: Same applies to a few other tests below

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfigTest.java
```diff
@@ -104,5 +192,60 @@ public void testStartHttpServer() {
         HttpServers.release(httpServerOptional.get());
         HttpServers.release(httpServerOptional2.get());
     }
-}
 
+    @Test
+    public void testStartHttpsServer() throws Exception {
```
mimaison, line 197: This test does not throw Exception

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/integration/ClientMetricsReporterHttpsIT.java
```diff
@@ -0,0 +1,177 @@
...
+import static io.strimzi.kafka.metrics.prometheus.http.SslTestUtils.trustingClientSslContext;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
+
+public class ClientMetricsReporterHttpsIT {
+
+    private static final int PORT = Listener.parseListener(ClientMetricsReporterConfig.LISTENER_CONFIG_DEFAULT).port;
+    private static final Duration TIMEOUT = Duration.ofSeconds(30L);
+    private static final String CONTAINER_CERTIFICATE = "/tmp/tls.crt";
+    private static final String CONTAINER_KEY = "/tmp/tls.key";
```
mimaison, line 51: Should we use `Files.createTempFile()` instead of hardcoding paths?

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/integration/ClientMetricsReporterHttpsIT.java
```diff
@@ -0,0 +1,177 @@
...
+        try (GenericContainer<?> producer = clientContainer(configs)
+                .withCopyFileToContainer(MountableFile.forHostPath(certificate), CONTAINER_CERTIFICATE)
+                .withCopyFileToContainer(MountableFile.forHostPath(key), CONTAINER_KEY)) {
+            producer.start();
+
+            verifyHttpsMetrics(producer);
+        }
+    }
+
+    private GenericContainer<?> clientContainer(Map<String, String> configs) {
```
mimaison, line 102: Can we avoid all this duplication and update the methods in `MetricsUtils` to handle SSL?

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/http/HttpServersTest.java
```diff
@@ -22,16 +22,16 @@ public class HttpServersTest {
     @Test
     public void testLifecycle() {
         Listener listener1 = Listener.parseListener("http://localhost:0");
-        HttpServers.ServerCounter server1 = HttpServers.getOrCreate(listener1, registry);
+        HttpServers.ServerCounter server1 = HttpServers.getOrCreate(listener1, registry, null);
```
mimaison, line 25: Why are we making these changes? We still have the `getOrCreate(Listener listener, PrometheusRegistry registry)` method

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java
```diff
@@ -38,7 +41,65 @@ public class ClientMetricsReporterConfig extends AbstractConfig {
...
+    /**
+     * Configuration key for the enabled cipher suites.
+     */
+    public static final String LISTENER_SSL_ENABLED_CIPHER_SUITES_CONFIG =
+            LISTENER_SSL_CONFIG_PREFIX + "enabled.cipher.suites";
+
+    /**
+     * Default value for the enabled cipher suites. An empty list uses Java's default cipher suites.
+     */
+    public static final String LISTENER_SSL_ENABLED_CIPHER_SUITES_CONFIG_DEFAULT = "";
```
mimaison, line 99: Do we really need a constant for the empty string? Why is it not `null` like some of the other files?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/HttpServers.java
```diff
@@ -33,9 +35,23 @@ private HttpServers() {
...
+     * @param listener The host and port
+     * @param registry The Prometheus registry to expose
+     * @param httpsConfigurator The HTTPS configurator, or {@code null} to use HTTP.
+     * @return A ServerCounter instance
+     */
+    public synchronized static ServerCounter getOrCreate(
+            Listener listener,
+            PrometheusRegistry registry,
+            @Nullable HttpsConfigurator httpsConfigurator) {
         ServerCounter serverCounter = SERVERS.get(listener);
```
mimaison, line 52: We could have the same `Listener` but different SSL configurations, in that case this would return an existing instance instead of an instance with the provided configurations

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java
```diff
@@ -60,17 +121,37 @@ public class ClientMetricsReporterConfig extends AbstractConfig {
...
+                    ConfigDef.Importance.HIGH, ALLOWLIST_CONFIG_DOC)
+            .define(LISTENER_ENABLE_CONFIG, ConfigDef.Type.BOOLEAN, LISTENER_ENABLE_CONFIG_DEFAULT,
+                    ConfigDef.Importance.HIGH, LISTENER_ENABLE_CONFIG_DOC);
 
     final Listener listener;
     final boolean listenerEnabled;
     final PrometheusRegistry registry;
     final Pattern allowlist;
+    SSLContextFactory sslContextFactory = null;
+    HttpsConfiguratorFactory httpsConfiguratorFactory = null;
```
mimaison, line 154: Want about making these `final` and setting them to `null` if needed in the constructor?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
```diff
@@ -30,54 +30,83 @@ public class Listener {
...
+
         @Override
         public void ensureValid(String name, Object value) {
-            parseListener(String.valueOf(value));
+            Matcher matcher = PATTERN.matcher(String.valueOf(value));
+            if (!matcher.matches()) {
+                throw new ConfigException(name, value, "The Listener must be of format http(s)://[host]:[port]");
+            }
```
mimaison, line 109: Why are we changing the logic of this method?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
```diff
@@ -120,8 +125,12 @@ public void contextChange(MetricsContext metricsContext) {
...
     }
 
-    // for testing
-    Optional<Integer> getPort() {
+    /**
+     * Gets the port used by the HTTP server.
+     *
+     * @return The HTTP server port, or empty if the HTTP server is not running.
+     */
+    public Optional<Integer> getPort() {
```
mimaison, line 133: Do we need to make this method `public`?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
```diff
@@ -58,8 +58,13 @@ public ClientMetricsReporter() {
...
 
-    // for testing
-    ClientMetricsReporter(PrometheusRegistry registry, KafkaCollector kafkaCollector) {
+    /**
+     * Constructor
+     *
+     * @param registry The Prometheus registry.
+     * @param kafkaCollector The Kafka metrics collector.
+     */
+    public ClientMetricsReporter(PrometheusRegistry registry, KafkaCollector kafkaCollector) {
```
mimaison, line 67: Do we need to make this `public`?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+
+import static io.strimzi.kafka.metrics.prometheus.ClientMetricsReporterConfig.LISTENER_SSL_CERTIFICATE_CONFIG;
+import static io.strimzi.kafka.metrics.prometheus.ClientMetricsReporterConfig.LISTENER_SSL_CERTIFICATE_LOCATION_CONFIG;
+import static io.strimzi.kafka.metrics.prometheus.ClientMetricsReporterConfig.LISTENER_SSL_KEY_CONFIG;
+import static io.strimzi.kafka.metrics.prometheus.ClientMetricsReporterConfig.LISTENER_SSL_KEY_LOCATION_CONFIG;
+
+/**
+ * Factory for building the SSL context used by the HTTPS listener.
+ */
+public class SSLContextFactory {
```
mimaison, line 40: I'm not very familiar with this type of logic and I've not looked at this file in depth yet but it looks more complicated than I expected.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+                    KeyManagerFactory.getDefaultAlgorithm());
+            keyManagerFactory.init(keyStore, keyPassword);
+
+            SSLContext sslContext = SSLContext.getInstance("TLS");
+            sslContext.init(keyManagerFactory.getKeyManagers(), null, SECURE_RANDOM);
+            return sslContext;
+        } catch (ConfigException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new ConfigException(
```
mimaison, line 105: We shouldn't pass an exception as the `value` argument in `ConfigException`

mimaison (commented):
Thanks for the PR. I made a quick first pass and left a few comments.
Also I think we should have an integration test in server-metrics-reporter

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/HttpServers.java
antonio-pedro99: We actually use this annotation to marke argument as nullable here https://github.com/strimzi/metrics-reporter/pull/130/changes#diff-f7dc806605b3fef56b97788f25dec8daae310c40fa17df51ca09d60555cd2c08R51

But this is the wrong argument, it should be maybe `org.jetbrains.annotations.Nullable`

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
antonio-pedro99: Yes, for testing.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java
antonio-pedro99: That's a good catch, does not really make sense.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java
antonio-pedro99: Sounds reasonable.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/http/HttpServersTest.java
antonio-pedro99: No, the method signature has been changed to accept a nullable HttpsConfigurator.
`@Nullable HttpsConfigurator httpsConfigurator`, since we do not need https in these tests, we pass null.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/integration/ClientMetricsReporterHttpsIT.java
antonio-pedro99: Yes, sounds like a good idea.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/integration/ClientMetricsReporterHttpsIT.java
antonio-pedro99: or maybe use the ` @TempDir
    private Path tempDir;` already in the code.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfigTest.java
antonio-pedro99: Because we deal with two factories in the assertions
- `sslContextFactory`
- `httpsConfiguratorFactory`

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfigTest.java
antonio-pedro99: Tbh, I need to revise this test. It suggests that we only set the listener and the rest configs should be the defaults, but there are some configs that are really needed according to the proposal.

README.md
```diff
@@ -161,4 +177,4 @@ If you want to get in touch with us first before contributing, you can use:
 - [GitHub Discussions](https://github.com/orgs/strimzi/discussions)
 
 ## License
-Strimzi is licensed under the [Apache License](./LICENSE), Version 2.0
\ No newline at end of file
+Strimzi is licensed under the [Apache License](./LICENSE), Version 2.0
```
antonio-pedro99, line 180: Unrelated change

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/HttpServers.java
mimaison: Yes I noticed you used this annotation in this file. My point is that the rest of the code base does not use this annotation.

scholzj: Just FYI: In other parts of Strimzi we intentionally avoid any @Nullable annotations.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/http/HttpServersTest.java
mimaison: Sorry but the `getOrCreate(Listener listener, PrometheusRegistry registry)` method is still there: https://github.com/antonio-pedro99/metrics-reporter/blob/5fc4d25ece064d10eac42f80e61e5f44e9221393/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/HttpServers.java#L37

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfigTest.java
mimaison: That wasn't clear when I reviewed the code

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/HttpServers.java
antonio-pedro99: thanks for informing @scholzj , I will get rid of it.

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfigTest.java
antonio-pedro99: should I resolve this?

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/http/HttpServersTest.java
antonio-pedro99: Fair, I will use that one instead of using the one that requires an HttpsConfigurator.

katheris:
Discussed on community call on 23.7.26: This hasn't had any recent activity, @antonio-pedro99 are you planning to address the comments from Mickael or should we close this PR if you aren't able to get back to it.

antonio-pedro99:
> Discussed on community call on 23.7.26: This hasn't had any recent activity, @antonio-pedro99 are you planning to address the comments from Mickael or should we close this PR if you aren't able to get back to it.

Hi, I will finish the PR. I can’t just find enough time in the last 3 weeks, hopefully I will be able to resume my contributions from next week

mimaison:
@antonio-pedro99 Do you have an update? If some any reason you can't finish this task, no worries, just let us know. Thanks

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
antonio-pedro99: i was getting a javadoc generation issue. It was asking for a constructor with comments. But I can revert that

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfigTest.java
antonio-pedro99: shhould I use singular word for this test's name?

antonio-pedro99:
> @antonio-pedro99 Do you have an update? If some any reason you can't finish this task, no worries, just let us know. Thanks

Hi, rlly sorry for taking this long. I am on it now.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java
antonio-pedro99: @mimaison , well this was intentional actually. empty strings are parsed to list. https://kafka.apache.org/37/javadoc/org/apache/kafka/common/config/ConfigDef.Type.html#:~:text=empty%20String%20maps%20to%20an%20empty%20List.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
antonio-pedro99: no, package-privated is enough

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
antonio-pedro99: @mimaison should you ask questions and I answer, or you want to propose a better logic?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
antonio-pedro99: I am not sure I understood, but seems like this is not an isolated case. I see [this](https://github.com/strimzi/metrics-reporter/blob/b689e746f675e1f6aa031db3c2d6944a0418a778/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfig.java#L113) uses the same pattern.

mimaison: This is calling the 2-argument constructor of `ConfigException`. That constructor expects a `value` (for the config) in the 2nd argument and this is passing an `Exception`, see https://kafka.apache.org/43/javadoc/org/apache/kafka/common/config/ConfigException.html

We should not pass exception to `ConfigException`

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporter.java
mimaison: The code seems to compile and run without it

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/ClientMetricsReporterConfigTest.java
mimaison: Yes I think you can resolve this

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/http/HttpServersTest.java
```diff
@@ -4,7 +4,7 @@
  */
 package io.strimzi.kafka.metrics.prometheus.http;
 
-import io.prometheus.metrics.model.registry.PrometheusRegistry;
+import  io.prometheus.metrics.model.registry.PrometheusRegistry;
```
mimaison: We don't need this extra space

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+            this.tag = tag;
+            this.valueOffset = valueOffset;
+            this.length = length;
+            this.endOffset = endOffset;
+        }
+    }
+
+    @Override
+    public String toString() {
+        return "SslContextFactory{" +
```
mimaison: nit: The class is named `SSLContextFactory` but this prints `SslContextFactory`

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+     */
+    public SSLContext create() {
+        try {
+            X509Certificate[] certificateChain = certificateChain(loadCertificateChain());
+            PrivateKey privateKey = privateKey(loadKey());
+            validatePrivateKeyMatchesCertificate(privateKey, certificateChain[0]);
+
+            char[] keyPassword = KEY_STORE_PASSWORD.toCharArray();
+
+            KeyStore keyStore = KeyStore.getInstance("PKCS12");
```
mimaison: Should we do `KeyStore.getInstance(KeyStore.getDefaultType())` to work in environments without PKCS12 support?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+        if (certificates.isEmpty()) {
+            throw new ConfigException(LISTENER_SSL_CERTIFICATE_CONFIG, null, "No X.509 certificates found");
+        }
+
+        X509Certificate[] certificateChain = new X509Certificate[certificates.size()];
+        int index = 0;
+        for (Certificate certificate : certificates) {
+            if (!(certificate instanceof X509Certificate)) {
+                throw new ConfigException(LISTENER_SSL_CERTIFICATE_CONFIG, null, "Only X.509 certificates are supported");
+            }
```
mimaison, line 163: Do we need this check? The `certificates` collection is built from `CertificateFactory.getInstance("X.509")`

mimaison: If we remove that check, we could also just return `certificates.toArray(new Certificate[0]);` and change the return type to `Certificate[]`.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+            try {
+                return Files.readString(Path.of(location), StandardCharsets.UTF_8);
+            } catch (IOException e) {
+                throw new ConfigException(
+                        locationConfig,
+                        location,
+                        "Failed to read " + description + ": " + e.getMessage());
+            }
+        }
+        throw new ConfigException(inlineConfig, null, description + " must be configured for HTTPS listeners");
```
mimaison: I don't think we should use the `ConfigException` 3-argument constructor if we pass `null`. In that case, we should probably just use the 1-argument constructor with just a message.

Same multiple times below

mimaison (commented):
Thanks for the updates, I made another pass and left a few more comments

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/Listener.java
ppatierno: We are doing the same across other Strimzi projects, because newer version of Java start complaining about that and printing the warnings. I think that they will be "promoted" to errors in the future.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    private String loadKey() {
+        return loadPem(
+                key,
+                keyLocation,
+                LISTENER_SSL_KEY_CONFIG,
+                LISTENER_SSL_KEY_LOCATION_CONFIG,
+                "SSL private key");
+    }
```
ppatierno, line 126: Tbh I am not sure we really need these two methods just to wrap `loadPem` call, taking also into account these two methods are called just one time and imho it's better to see what they are doing.
I would just remove them and use `loadPem`.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+                throw new ConfigException(
+                        LISTENER_SSL_KEY_CONFIG,
+                        null,
+                        "Failed to parse private key: " + e.getMessage());
+            }
+        }
+        throw new ConfigException(LISTENER_SSL_KEY_CONFIG, null, "Failed to parse private key");
+    }
+
+    private PrivateKey parsePrivateKey(String algorithm, byte[] keyBytes, String keyType) {
```
ppatierno, line 201: I don't understand the difference between `algorithm` and `keyType` even because this method is always called with them equals.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+            offset = value.endOffset;
+        }
+
+        throw new ConfigException(
+                LISTENER_SSL_KEY_CONFIG,
+                null,
+                "EC private key must include named curve parameters");
+    }
+
+    private void validatePrivateKeyMatchesCertificate(PrivateKey privateKey, X509Certificate certificate) {
```
ppatierno, line 277: Why do we need this complex logic here? I am not sure it's responsability of the metrics reporter to check that the user has provided the private key which matches the certificate. If they don't match, the TLS handshake will fail anyway. We don't have such check in any Strimzi component, not sure we need it here.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
```diff
@@ -0,0 +1,459 @@
...
+            if (!(certificate instanceof X509Certificate)) {
+                throw new ConfigException(LISTENER_SSL_CERTIFICATE_CONFIG, null, "Only X.509 certificates are supported");
+            }
+            certificateChain[index] = (X509Certificate) certificate;
+            index++;
+        }
+        return certificateChain;
+    }
+
+    private PrivateKey privateKey(String pem) {
```
ppatierno, line 170: From this method, several other methods are implemented because we are trying to handle formats which are different from PKCS8 and we try to wrap them into a PKCS8 to be used in the end within the `PKCS8EncodedKeySpec`. Is this complexity really needed? The PKCS8 format is the most widely used so I don't think we need such logic for other formats. It would simplify the class a lot, by removing maybe 60% of the code? @mimaison wdyt?

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
ppatierno: I also think this class is bringing complexity we don't need. I left some comments.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
ppatierno: Also with `signatureAlgorithm` you are hardcoding signatures with 256 bits, but it could be 512 or less so the code would not work at all.

ppatierno (changes requested):

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
mimaison: As I said in another comment, I wasn't expecting this much logic and complexity. But I'm not sure if it's needed by Strimzi or not.

ppatierno: When users provide a private key and certs for listeners within Strimzi (https://strimzi.io/docs/operators/latest/configuring#property-listener-config-brokerCertChainAndKey-reference) they have to be in PKCS8 format. I think we could have the same here in the metrics reporter and leaving the user the need to convert any other format to PKCS8 instead of this really complex logic here.

ppatierno: Let me also add that while PKCS8 is a generic format (actually wrapping the others) to store different type of private keys, the PKCS1 is RSA specific as well as SEC1 is EC specific (even the header highlights it) . When it will come to ML-DSA, the PKCS8 format will be likely to be used because generic enough to host different kind of private keys. So I would just stick with PKCS8.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/HttpServers.java
antonio-pedro99: @mimaison I will come with a better design, allow me some time

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
antonio-pedro99: @ppatierno thanks for the review, I will maybe propose a better implementation and address the comments. I addressed some comments from @mimaison today.

client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/http/SSLContextFactory.java
antonio-pedro99: > The certificates collection is built from CertificateFactory.getInstance("X.509")

I did not know that, in this case, the check is not needed.

antonio-pedro99:
@mimaison @ppatierno 

I have been thinking about this PR for a while.

Comparing to how other Strimzi components handle TLS encryption, this component has no way to get rid of some complexities presented in this PR. 

The Kafka Bridge uses vert.x which provides built-in support to handle TLS/SSL and PEM certs. This is different from this component as we expose metrics through prometheus [HttpServer](https://prometheus.github.io/client_java/exporters/httpserver/). The prometheus's httpServer does not provide any built-in mechanism to handle TLS/SSL materials, they just provide a way to configure a HTTPS through HttpConfigurator. The configurator expects SSLContext which will need to be initialized with the configurations we provide as listener ssl configuration parameters. We will need to load and parse file or inline PEMs on our own.

One solution would be to expose the metrics through vert.x http server. In the context of adding only support for tls encryption, replacing prometheus httpServer with vert.x seems too much, but it will help us avoid making the metrics-reporter owning abstractions it does not need to handle(load, parse and validate PEM and etc). 

This is my proposal, and I am open for suggestions.

ppatierno:
I don't think we should add an additional dependency like Vert.x just for making SSL configuration simpler.
I think that all the logic within the `SslContextFactory` would be simplified a lot if we narrow the support to PKCS#8. All the complex part around dealing with DER format, PKCS#1 etc will go away. Any RSA or EC key can be put inside a PEM file with PKCS#8 format. The PKCS#1 and SEC1 are specific to RSA and EC but they are not needed at all imho.

antonio-pedro99:
yes, I agree.
I narrowed the SslContextFactory to handle PCKS#8 a week ago; however, I'm still not convinced by the approach. I will update the PR for further review.

PR #131, opened 2026-05-30
Run integration tests against latest patch Kafka versions

guancioul:
Close #107,

### Summary

Integration tests currently run against a single hardcoded Kafka version. This PR runs them against all latest patch versions of each supported Kafka minor (e.g. 3.8.x, 3.9.x, 4.0.x) to catch version-specific regressions automatically.

- Upgrade JUnit from 5.10.2 to 6.0.0 to use `@ParameterizedClass`, which parameterizes an entire test class per argument set
- Bump strimzi-test-container from 0.115.0 to 0.116.0, which introduces the `StrimziKafkaCluster#getLatestPatchVersions()` factory method used as the parameter source

### Note

This PR depends on https://github.com/strimzi/test-container/pull/226 (merged, releasing in 0.116.0).
The implementation follows the same approach used in https://github.com/strimzi/strimzi-kafka-bridge/pull/1053.

### DIFF
--- a/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
+++ b/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
@@ -43,7 +43,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 public class MetricsUtils {
 
     public static final String VERSION = "1.0.0-SNAPSHOT";
-    private static final String CLIENTS_IMAGE = "quay.io/strimzi-test-clients/test-clients:0.12.0-kafka-4.0.0";
+    private static final String CLIENTS_IMAGE_PREFIX = "quay.io/strimzi-test-clients/test-clients:latest-kafka-";
     private static final Duration TIMEOUT = Duration.ofSeconds(30L);
 
     public static final String REPORTER_JARS = "target/client-metrics-reporter-" + VERSION + "/client-metrics-reporter-" + VERSION + "/libs/";
@@ -177,10 +177,11 @@ public class MetricsUtils {
      * Start a test-clients container
      * @param env the environment variables
      * @param port the port to expose
+     * @param kafkaVersion the Kafka version the client should run against
      * @return the container instance
      */
-    public static GenericContainer<?> clientContainer(Map<String, String> env, int port) {
-        return new GenericContainer<>(CLIENTS_IMAGE)
+    public static GenericContainer<?> clientContainer(Map<String, String> env, int port, String kafkaVersion) {
+        return new GenericContainer<>(CLIENTS_IMAGE_PREFIX + kafkaVersion)
                 .withNetwork(Network.SHARED)
                 .withExposedPorts(port)
                 .withCopyFileToContainer(MountableFile.forHostPath(REPORTER_JARS), MOUNT_PATH)
--- a/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
+++ b/client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
@@ -43,7 +43,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 public class MetricsUtils {
 
     public static final String VERSION = "1.0.0-SNAPSHOT";
-    private static final String CLIENTS_IMAGE_PREFIX = "quay.io/strimzi-test-clients/test-clients:latest-kafka-";
+    private static final String CLIENTS_IMAGE = "quay.io/strimzi-test-clients/test-clients:0.14.0-kafka-4.2.0";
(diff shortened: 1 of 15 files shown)


see-quick (changes requested):
Hi @guancioul, thanks for the PR. But looking to this again ... I think we need look at this from a higher level.

The core goal of this PR is to verify that metrics reporter works corretly acrros different versions. But right now, the kafkaVersion parameter is only being applied to the Kafka broker cluster, not the components that are actually under test. In `TestConnectMetricsIT` and `TestMirrorMakerMetricsIT`, the `StrimziConnectCluster` is never configured with .withKafkaVersion(kafkaVersion) .. so that means Connect workers always run with default version (i.e., latest one). 

Similarly in `TestConsumerMetricsIT`, `TestProducerMetricsIT` and `TestStreamsMetricsIT` the client container is created from the hardcoded image `test-clients:0.12.0-kafka-4.0.0` so basically consumer, producer and streams app always run on Kafka 4.0.0 regardless what version is parametrized. 

Also if we think about this more deeply ... then we can realise that it doesn't make much value if you are in `TestConnectMetricsIT` and you are also checking the broker version instead of directly just overriding the Connect workers. Same applies for Producer, Consumer, Streams etc.  

So we need to address these issues first.

guancioul:
Thanks for pointing this out! @see-quick. My original change only wired `kafkaVersion` into the Kafka broker cluster; I didn't realize it never actually propagated to the components under test, so the producer/consumer/streams clients and the Connect workers kept running on a fixed/default Kafka version regardless of the parametrized broker version. That meant the PR wasn't really testing what it claimed to.

I've fixed this so every component now picks up the same `kafkaVersion`, and I verified locally against both latest patch versions (4.2.1 and 4.3.0) — all tests pass. I confirmed in the logs that the broker/Connect worker/client container all consistently use the same version per run.

see-quick:
Thanks for the update @guancioul, the version propagation fix looks good!

One more thing ... I think we should only parameterize the version of the component that is actually the SUT, not the Kafka broker cluster across all tests. We are copying the Metrics Reporter plugin into a specific component and verifying it there, so that is what we should be varying. The broker is just supporting infrastructure in most tests but I think I am also fine with setting it the same version ... but it doesn't add any value here if you understand me. It is just un necessary noise. 

Any more opinions on this @strimzi/maintainers or @mimaison ?

katheris:
Discussed on community call on 23.7.26: We decided that the suggestion made by @see-quick above makes sense, so @guancioul can you please update this PR to follow that suggestion?

guancioul:
Thanks @see-quick and @katheris, and sorry for the slow follow-up, I have been occupied with some interviews recently. Updated per the community call, I removed the `kafkaVersion` from the supporting brokers, to keep the code cleaner and reduce noise.

see-quick (approved):
Now it looks good thanks 👍 !

client-metrics-reporter/src/test/java/io/strimzi/kafka/metrics/prometheus/MetricsUtils.java
```diff
@@ -43,7 +43,7 @@
 public class MetricsUtils {
 
     public static final String VERSION = "1.0.0-SNAPSHOT";
-    private static final String CLIENTS_IMAGE = "quay.io/strimzi-test-clients/test-clients:0.12.0-kafka-4.0.0";
+    private static final String CLIENTS_IMAGE_PREFIX = "quay.io/strimzi-test-clients/test-clients:latest-kafka-";
```
im-konge: Is it right to change it like this? What will happen in case that there are some new versions of Kafka that are reflected in `latest` tag of test-clients, older versions are removed, but the test-containers doesn't have updated matrix of Kafka versions? I mean, in case that you will use some other version than supported by the `latest` tag (or `main` branch) of test-clients, the tests will simply fail. Does it matter that we are testing with different versions of clients?

im-konge: I know that you had comment from @see-quick about changing it to be parametrized, but I think it will create issues in the future.

guancioul: Thanks for pointing this out. Using parametrized client images could cause issues when test-clients and test-container versions get out of sync. For now, I think switching back to a pinned image, `0.14.0-kafka-4.2.0`, is the most stable option, even though it reduces client version coverage. I'm not sure there's a better approach that keeps both stability and multi-version coverage. What do you think? @see-quick @im-konge

im-konge: Yeah I don't think there is a different way how to handle that right now.

guancioul: Thanks @im-konge. I've updated the client ITs to use the pinned image `0.14.0-kafka-4.2.0` and removed the client-side version parametrization. Pls take a look!

PR #136, opened 2026-08-31
Bump Prometheus to 1.7.0 and fix label deduplication for the new API

ppatierno:
### Type of change

- Dependency update

### Description

This PR bumps the Prometheus Java client to 1.7.0 to align with latest update on the Strimzi operator and HTTP bridge. A plain version bump is not enough because this release changed the contract of `PrometheusNaming.sanitizeLabelName()`: it no longer replaces illegal characters such as hyphens with underscores. Such a normalisation was moved into `prometheusName()`, which is now called internally by Labels.of() before validation.

Our duplicate-label guard in `KafkaMetricWrapper`, `YammerMetricWrapper`, and `DataPointSnapshotBuilder` was using `sanitizeLabelName()` as the dedup key. When a metric carries two tag keys that differ only by a hyphen vs. underscore (e.g. k-1 and k_1), the old method mapped both to k_1 so the HashSet caught the collision. With 1.7.0 it returns k-1 unchanged, both keys pass the guard, and Labels.of() then throws IllegalArgumentException: k_1: duplicate label name when it detects the collision itself.

The fix is is using `prometheusName(sanitizeLabelName(key))` instead, which mirrors what the library now does internally and restores correct dedup behaviour.

### Checklist

- [x] Make sure all tests pass
- [x] AI assistance was used to create this PR (see the [Strimzi AI policy](https://github.com/strimzi/governance/blob/main/AI_POLICY.md))

### DIFF
--- a/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/common/DataPointSnapshotBuilder.java
+++ b/client-metrics-reporter/src/main/java/io/strimzi/kafka/metrics/prometheus/common/DataPointSnapshotBuilder.java
@@ -33,7 +33,7 @@ public class DataPointSnapshotBuilder {
      * @return The {@link InfoSnapshot.InfoDataPointSnapshot} datapoint
      */
     public static InfoSnapshot.InfoDataPointSnapshot infoDataPoint(Labels labels, Object value, String metricName) {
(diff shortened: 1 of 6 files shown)


mimaison (approved):
LGTM

